PR #1, opened 2015-09-11
Corrected translation of "Send Key"

bwildenhain:
Menu item "Send key" got translated to "Verstuur sleutel" ("sleutel
means "key" as in encryption), corrected to "Verstuur toets".


crobinso:
Thanks, commit is pushed now. FWIW we handle translations using fedora.zanata.org, so please consider signing up there if you notice more typos, it's a bit more work for me to integrate fixes locally and push them back to the server. https://github.com/virt-manager/virt-manager/blob/master/HACKING has a bit more info

PR #2, opened 2015-10-01
Added support pvspinlock in features

Akasurde:
Signed-off-by: Abhijeet Kasurde <akasurde@redhat.com>


crobinso:
Thanks Abhijeet, I applied manually and pushed it. However for any patches that are more complex than typo fixes, please continue to send them to the mailing list where more people are watching.

PR #3, opened 2015-10-02
addstorage: Fix a typo in an error message

cgwalters:
Just noticed it while using the app.


crobinso:
Thanks Colin, I applied your patch manually and pushed it

PR #4, opened 2015-12-09
Fixed Vte version to be compatible with Ubuntu 14.04

astrohome:
This will fix an error in Ubuntu 14.04 as only Vte 2.90 is available.


crobinso:
Thanks for the patch, but I actually went a different route and just dropped the require_version entirely, since virt-manager can work fine with both the 2.90 and 2.91 API versions.

PR #7, opened 2016-02-23
Suport port number in FTP urls

int-0:
Current version of _virt-manager_ doesn't handle FTP urls correctly: _"ftp://server:port/"_ causes exception because the **FTP()** constructor only support hostname, without port definition. To connect to a different port the only way is with **FTP().connect()** method.

### DIFF
--- a/virtinst/urlfetcher.py
+++ b/virtinst/urlfetcher.py
@@ -214,7 +214,12 @@ class _FTPURLFetcher(_URLFetcher):
 
         try:
             server = urlparse.urlparse(self.location)[1]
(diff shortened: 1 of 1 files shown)


crobinso:
Thanks for the bug report and the patch!

A couple suggestions:
- You can get the port value from the result of urlparse. o = urlparse.urlparse(...); print o.hostname, o.port
- Then we can unify the ftp init path and always call connect(hostname, port)

crobinso:
I pushed a patch with my suggested changes

PR #8, opened 2016-03-24
Disable x2apic for solaris 11

ariscop:
Same hack that 10 needs


ariscop:
https://liquidat.wordpress.com/2016/01/08/howto-solaris-11-on-kvm/

crobinso:
Thanks! Applied manually and pushed

PR #9, opened 2016-03-29
Make SpiceClientGtk optional

agx:
Not all architectures supported by Debian have it. This will allow to
make VNC work nevertheless.

### DIFF
--- a/virtManager/console.py
+++ b/virtManager/console.py
@@ -29,7 +29,7 @@ from .baseclass import vmmGObjectUI
 from .details import DETAILS_PAGE_CONSOLE
 from .serialcon import vmmSerialConsole
 from .sshtunnels import ConnectionInfo
-from .viewers import SpiceViewer, VNCViewer
+from .viewers import SpiceViewer, VNCViewer, have_spice_gtk
 
 
 # console-pages IDs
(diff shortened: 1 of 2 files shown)


crobinso:
Thanks, pushed. It's not exhaustive since virt-install and virt-manager will likely still try and default to spice, but I'd prefer to ignore going down that route if we can avoid it.

Also note, Fedora has kinda worked around this. spice doesn't build on all archs, but spice-gtk _is_ buildable on all archs by moving spice-protocol into its own package IIRC, since that's the only bit spice-gtk needs and it is arch indepdendent

agx:
On Thu, Apr 07, 2016 at 01:15:17PM -0700, Cole Robinson wrote:
> Thanks, pushed. It's not exhaustive since virt-install and
> virt-manager will likely still try and default to spice, but I'd
> prefer to ignore going down that route if we can avoid it.

Thanks a lot! I think people were most annoyed that virt-manager is not
even starting with spice msissing. Changing the display to VNC manually
should be fine for the moment.

> Also note, Fedora has kinda worked around this. spice doesn't build on all archs, but spice-gtk _is_ buildable on all archs by moving spice-protocol into its own package IIRC, since that's the only bit spice-gtk needs and it is arch indepdendent

Thanks for the hint!
 -- Guido

PR #10, opened 2016-06-17
virt-install: added a nousb device-option for suppress adding usb-devices

knumskull:
I needed this options for creating virtual machines without having usb-devices by default.


crobinso:
This isn't necessarily sufficient, since libvirt will also add a usb1 controller by default. Does --controller type=usb,model=none do what you want?

knumskull:
Yes. will work. Doesn't know the model=none option. You can reject my request

PR #11, opened 2016-06-22
Update de.po

Atalanttore:
Updated german translation for virt-manager.


Antique:
Hi, thanks for contributing to virt-manager.  We use zanata to handle translations <https://fedora.zanata.org/project/view/virt-manager>, would you be able to add those translation there?

crobinso:
Yes as stated above our translations our handled via zanata, and it's kind of difficult to try and merge manual changes outside of zanata. If you would like to help with translation please sign up there

PR #12, opened 2016-06-29
Update it.po to fix variable translation

andreamtp:
Fixed also on Zanata, related to: https://bugzilla.redhat.com/show_bug.cgi?id=1350185


crobinso:
There was another issue like this is in the file. I grabbed updated bits from zanata and fixed the broken instances manually

PR #13, opened 2016-09-02
Fix typo: vachine -> machine

yurchor:
Thanks for fixing this typo.


crobinso:
Thanks, rebased in and pushed now

PR #16, opened 2017-01-05
Add caveats section to virt-xml.pod

dhgutteridge:
Caveats section added to warn users that some changes made to a VM's definition may be ignored by the virtualization host.


crobinso:
Sorry for the delay, I've rebased this patch on top of master and pushed now

PR #17, opened 2017-01-24
Corrected translation for zh_CN

Arondight:
Remove the '%s', it is strange here.


crobinso:
Thanks for the patch. Translations are typically handled at https://fedora.zanata.org/project/view/virt-manager/  but since this is a simple change I've rebased it in and pushed now.

PR #15, opened 2016-12-26
Markdown format

mjtrangoni:
Hi @crobinso, 
I converted INSTALL, HACKING, and NEWS files to the Markdown format as I mentioned in [#14](https://github.com/virt-manager/virt-manager/pull/14). You can merge it if you like it.

PR #18, opened 2017-02-12
Markdown format

mjtrangoni:
Hi @crobinso, 

I made all changes like you suggest in [#15](https://github.com/virt-manager/virt-manager/pull/15) about the .spec and MANIFEST, and also converted the file COPYING. Now are all files in Markdown format.

A `python setup.py rpm` was successful for me on Fedora 25.


crobinso:
Thanks. I dropped the COPYING change: that's a boilerplate file that I'd rather just leave as is. The RPM/manifest changes I squashed into the individual commits. Pushed now, thanks again!

PR #19, opened 2017-02-14
Remove minor redundancy in action_remove_device()

dhgutteridge:
There's no need to call util.listify() twice.


dhgutteridge:
A trivial code tweak/micro-optimization.

crobinso:
Thanks, pushed now

PR #20, opened 2017-03-21
Fix ambigous label on Autostart checkbox

thurstylark:
When the Autostart checkbox on the host Virtual Network or Storage tabs
is deselected, the label reads "Never", and when selected the label reads
"On Boot". This changes these labels to always read "On Boot".


crobinso:
Yeah I agree that's confusing. Thanks for the patch, pushed now

PR #21, opened 2017-03-26
Update virt-install man with path to Fedora 25 ISO

mykaul:
The path was to an outdated Fedora 21. Updated for Fedora 25.


crobinso:
Thanks, pushed now

PR #23, opened 2017-04-15
This is now removed in patches on Ubuntu, so no longer necessary.

BryanQuigley:
virt-manager-1.3.2/debian/patches/virtinst/use_qxl_for_ubuntu.patch

That patch removing these lines is also from Marc Deslauriers on 2016-01-28

Start of my project to get more Ubuntu/Debian patches upstream (related to https://bugs.launchpad.net/ubuntu/+source/virt-manager/+bug/1667114).

### DIFF
--- a/virtinst/osdict.py
+++ b/virtinst/osdict.py
@@ -516,21 +516,6 @@ class _OsVariant(object):
         if guest.os.is_pseries():
             return "vga"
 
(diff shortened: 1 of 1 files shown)


crobinso:
Thanks, I tweaked the commit message a bit and pushed now

PR #26, opened 2018-03-28
support username and password in prepareLocation

hamzy:
If --location is an ftp url with a username and password
then virt-install fails to install with an error:
ERROR Error validating install location: Opening URL u failed: 530 Login incorrect..

### DIFF
--- a/virtinst/urlfetcher.py
+++ b/virtinst/urlfetcher.py
@@ -207,13 +207,20 @@ class _FTPURLFetcher(_URLFetcher):
             return
 
         try:
-            parsed = urllib.parse.urlparse(self.location)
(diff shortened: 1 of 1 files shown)


rst0git:
Hi @hamzy would you be able to your patches to the mailing list virt-tools-list@redhat.com?
You can do that with `git send-email`. See [HACKING.md](https://github.com/virt-manager/virt-manager/blob/master/HACKING.md#submitting-patches) for more information.

crobinso:
the patch was sent upstream, we will continue to discuss there: https://www.redhat.com/archives/virt-tools-list/2018-April/msg00000.html

closing this

PR #27, opened 2018-03-29
virt-xml: Warn when --update is inapplicable

dhgutteridge:
Provide a warning if a user supplies --update on a VM that's inactive, as that may be contrary to their expectations (unintended option use, unexpected VM outage, or the utility's failure to retrieve a reference to a VM that is in fact running).

### DIFF
--- a/virt-xml
+++ b/virt-xml
@@ -430,9 +430,13 @@ def main(conn=None):
             print_stdout(dev.get_xml_config())
         return 0
 
-    if options.update and active_xmlobj:
-        devs, action = prepare_changes(active_xmlobj, options, parserclass)
-        update_changes(domain, devs, action, options.confirm)
+    if options.update:
+        if active_xmlobj:
+            devs, action = prepare_changes(active_xmlobj, options, parserclass)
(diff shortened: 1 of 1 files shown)


rst0git:
Hi @dhgutteridge would you be able to your patches to the mailing list virt-tools-list@redhat.com?
You can do that with git send-email. See [HACKING.md](https://github.com/virt-manager/virt-manager/blob/master/HACKING.md#submitting-patches) and https://libvirt.org/hacking.html#patches for more information.

dhgutteridge:
Hi @rst0git,

I can certainly send this patch to the virt-tools email list if it's useful, but I noted HACKING.md states "Small patches are acceptable via github pull-request, but anything non-trivial should be sent to the mailing list." This patch I've submitted is trivial (from my perspective, anyway). Has this definition changed? I intend to send more small patches later, e.g. a one line patch to a man page.

crobinso:
No the policy hasn't changed, I think Radostin was just trying to be helpful and redirect to the list where most patches are discussed.

Regardless, I tweaked the print to logging.warn() which will show up on the console too, and pushed. Thanks for the patch!

PR #28, opened 2018-04-06
virt-xml: note --update implies --define

dhgutteridge:
A minor man page tweak for clarity.


crobinso:
Tweaked the subject and pushed, thanks!

PR #29, opened 2018-04-08
setup: Install files in new virtinst folders

z3ntu:
In commits fe9ed23 and 3909c10 several files were moved to subfolder of
virtinst, but they are not covered by the existing glob in the setup.py
file. Fix that by adding those subfolders explicitely.


rst0git:
Why explicitly adding `virtinst/devices/*.py` and `virtinst/domain/*.py`?

The following will avoid similar issues in future.
```python
("share/virt-manager/virtinst", glob.glob("virtinst/*/*.py"))
```

PR #32, opened 2018-09-26
connection: avoid failing on NoneType in _hostinfo

cpaelzer:
There is a potential race of a backend to be opened and being true on
self._backend.is_open but self._hostinfo not being set yet.

Avoid spurial bugs due to that by also checking against None before
accessing subelements ot self._hostinfo

### DIFF
--- a/virtManager/connection.py
+++ b/virtManager/connection.py
@@ -318,12 +318,12 @@ class vmmConnection(vmmGObject):
     caps = property(lambda self: getattr(self, "_backend").caps)
 
     def host_memory_size(self):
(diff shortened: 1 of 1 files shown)


cpaelzer:
Found this while debugging [Ubuntu bug 1794315](https://bugs.launchpad.net/ubuntu/+source/virt-manager/+bug/1794315).
While I can't elaborately show the race in all its complexity I think the safety check makes sense to avoid such issues from crashing the tool.

crobinso:
Makes sense to me, I've pushed your patch now. Thanks!

PR #33, opened 2018-10-09
Add missing argument in message translation

kgawda:
Add missing "--location" in PL translation of "CDROM media does not print..."


crobinso:
Thanks for the patch, I've merged it now. However please use https://fedora.zanata.org for future translation changes, thats the central tool we use for translations

PR #34, opened 2018-10-09
virtinst: quickfix ubuntu net-preseed insert cdrom error

JOJ0:
The proper kernel/initrd pair for booting from an http server is found in install/netboot/... on a mounted ubuntu iso. This last worked in v1.4.2

I am sure this is not a proper fix, please point me in the right direction.

I am using virt-install to preseed Ubuntu machines from an http server. The issue is that in urldetect.py the media_type is identified by this

```
 elif cache.content_regex(".disk/info",
                "%s.*" % cls._debname.capitalize()):
            media_type = "disk"
```

Therefore if the install source where kernel/initrd is pulled from is an http server with the mounted ubuntu iso, it gets identified as a normal local disk install and the wrong kernel is loaded: install/vmlinuz or something that is decided in:

`_set_installcd_paths(self)`

It should rather jump to the network install method:

 `def _set_url_paths(self):`

See the commit for how I quickfixed tis.

I suppose the better approach would be to also check here in line 687 wether we this url is definitely from a network server and therefore should be prefered instead of the kernel for install directly from a cdrom:

```
       elif cache.content_regex("install/netboot/version.info",
                "%s.*" % cls._debname.capitalize()):
            media_type = "mounted_iso_url"

```

In 1.4.2 it just worked like this:

class UbuntuDistro(DebianDistro):
-    # http://archive.ubuntu.com/ubuntu/dists/natty/main/installer-amd64/
-    name = "Ubuntu"
-    urldistro = "ubuntu"
-
-    def isValidStore(self):
-        if self.fetcher.hasFile("%s/MANIFEST" % self._url_prefix):
-            # For regular trees
-            filename = "%s/MANIFEST" % self._url_prefix
-            regex = ".*%s.*" % self._installer_dirname
-        elif self.fetcher.hasFile("install/netboot/version.info"):
-            # For trees based on ISO's
-            self._url_prefix = "install"
-            self._set_media_paths()
-            filename = "%s/netboot/version.info" % self._url_prefix
-            regex = "%s*" % self.name
-        elif self.fetcher.hasFile(".disk/info") and self.arch == "s390x":
-            self._hvm_kernel_paths += [("boot/kernel.ubuntu", "boot/initrd.ubuntu")]
-            self._xen_kernel_paths += [("boot/kernel.ubuntu", "boot/initrd.ubuntu")]
-            filename = ".disk/info"
-            regex = "%s*" % self.name
-        else:
-            return False

Thanks for any hints for getting this fixed properly.

all the best!
Jojo

### DIFF
--- a/virtinst/urldetect.py
+++ b/virtinst/urldetect.py
@@ -684,6 +684,9 @@ class DebianDistro(Distro):
             media_type = "url"
         elif check_manifest("daily/MANIFEST"):
             media_type = "daily"
+        elif cache.content_regex("install/netboot/version.info",
+                "%s.*" % cls._debname.capitalize()):
+            media_type = "mounted_iso_url"
         elif cache.content_regex(".disk/info",
                 "%s.*" % cls._debname.capitalize()):
             media_type = "disk"
@@ -732,6 +735,9 @@ class DebianDistro(Distro):
         url_prefix = "current/images"
         if self.cache.debian_media_type == "daily":
             url_prefix = "daily"
+
+        if self.cache.debian_media_type == "mounted_iso_url":
+            url_prefix = "install"
 
         self._boot_iso_paths = ["%s/netboot/mini.iso" % url_prefix]
 


JOJ0:
Hi, thanks. This one: 
http://cdimages.ubuntu.com/ubuntu/releases/bionic/release/ubuntu-18.04.1-server-amd64.iso

JOJ0:
Hi again,
some more hints a came across along the way:

The first ISO image you find when searching for latest ubuntu lts isos is the live-server image:
http://releases.ubuntu.com/18.04/

It is stripped down and _not_ intended for network installs, don't try it! (the install/netboot/ directory is missing at all)

the issue described above does also happen if you use 16.04 or 14.04 server isos. the folder structure install/netboot/... is the same/similar.

hope that helps
keep me posted :-)
Jojo

crobinso:
Okay I just played with this a bit. Current virt-manager.git-fetched kernels boot fine into the text installer, but the kernel/initrd found by your patch boot fine too. I didn't go further because I don't really know what I'm looking for.

Can you explain a bit more how the behavior is different between the two kernels? How are you trying to use them exactly, how is the current state failing, and what does a success look like after your patch?

I just want to make sure that I get the details right so I can understand the case in the future so this doesn't regress

JOJ0:
Hi,
thanks for checking it out and getting back to me.

We do automated installs as follows:

ISO is mounted and its contents shared via http on a server. On one of my libvirt hypervisors if I let virt-install boot from the default kernel on the iso, the installer starts but I pretty soon get this error in the Ubuntu installer and it can't continue:

"Your installation CD-ROM couldn't be mounted. This probably means that the CD-ROM is not in the drive. If yo you can insert it and try again. Retry mounting the CD-ROM? yes/no"

In the syslog of the installer I see that at installer menu-item "cdrom-detect", the error happens. Looks like some variable is empty because no cdrom device can be found.

In short: I suppose the default vmlinuz/initrd combo is not intended for installs where the mounted cdrom is provided via a remote url. That's why the netboot/... combo is there.

If I use my fix and boot with the netboot/... kernel I don't get this message and install is fine.

My fix lacks one thing because I couldn't figure it out yet, I suppose you could help me with that:

We are checking if netboot/... is there, if it is, it is just used and it doesn't come to the next condition where it sould check if .disk/version is there....So what if a user uses this kind of iso but doesn't want to do a preseed install from network, but do a local preseed install from cdrom. Probably (I didn't test it) he would need the default kernel which includes the detect-cdrom part.

So my question would be: How can we figure out at that our iso source is definitely a network url and we should better decide for the netboot/ kernels.

At this point (line 688) we should add something to the condition like: "AND is_a_network_url()"

https://github.com/virt-manager/virt-manager/pull/34/commits/8f4f6932f20df62ca32c2f4f159c5b8e065c3fe1#diff-3779559bb64632a19049773bff2fb9eeR687

Is there some property of some object available at this time where I could get this information? Or am I a missing something and at this point we already are definitely are sure that we are installing from a network source and not from a local cdrom/iso mount?

crobinso:
Sorry I had to prioritize some other stuff, I'll be looking at this today or tomorrow

JOJ0:
Hi Cole, I have to thank you! I did some test-installation and it works fine :-)
all the best!
Jojo

PR #37, opened 2018-11-01
Update create.py

kalebskeithley:
100 is a bit too short for managing/creating vms over the corporate VPN. About half the time it times out and I have to exit the virtmgr app and restart it to see the new vm.


crobinso:
Thanks for the patch, pushed now

PR #39, opened 2018-12-24
urldetect: add support for old-style distros

yuvalturg:
Some distros do not have a .treeinfo file, but the location for the
kernel and initrd is fixed under images/pxeboot/{vmlinuz,initrd.img}.
Added a GenericDistro class as the last class in the store to support
this distro style.

Signed-off-by: Yuval Turgeman <yturgema@redhat.com>

### DIFF
--- a/virtinst/urldetect.py
+++ b/virtinst/urldetect.py
@@ -41,6 +41,9 @@ class _DistroCache(object):
             self._filecache[path] = content
         return self._filecache[path]
 
+    def paths_exist(self, paths):
+        return all([self._fetcher.hasFile(p) for p in paths])
+
     @property
     def treeinfo(self):
         if self._treeinfo:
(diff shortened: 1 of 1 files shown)


yuvalturg:
Yes, there's no treeinfo file in the ovirt-node-ng installer iso, but it is missing in Fedora and CentOS NetInstall ISOs as well (the ovirt-node-ng iso is derived from the netinstall isos)

crobinso:
I'm trying to avoid reviving this behavior if I can. There have been similar reports to yours though. We might be able to leverage libosinfo here which would be better, I'm planning to explore it.

One more question, are you using virt-install --location $ISO, or mounting+exporting the iso contents over HTTP and doing virt-install --location $URL ?

crobinso:
Thanks for the code but I tried to address this in a different way upstream. We now ask libosinfo for help identifying iso kernel/initrd locations. It uses a different database and can recognize many boot.iso style media. For example it detects ovirt-node-ng-installer-4.2.0-2019012210.el7.iso as RHEL7, which also has the added benefit that virt-install doesn't require an explicit --os-variant which is easy to miss and without it you'll get crap non-virtio defaults.

Additionally I added some CLI support for manually pointing at kernel/initrd paths in an ISO if virt-install fails to detect it, so there's a bit of a workaround. Say foo.iso isn't detected by libosinfo/virt-install, but it has a kernel at images/vmlinuz and initrd at images/initrd.gz, you can do

  virt-install --location foo.iso,kernel=images/vmlinuz,initrd=images/initrd.gz

So there's at least a workaround. But if you find media that libosinfo/virt-install doesn't detect, please file a libosinfo bug.

yuvalturg:
That is great, I will check it :+1:
Thanks !

PR #40, opened 2019-01-19
virt-install.pod: make example command match commentary

kenyon:
The example commentary says Fedora 20, so make the command line match that.


Antique:
Hi, thanks for the patch. Merged now.

PR #42, opened 2019-02-04
Complete strings not acessible on Zanata

p-bo:
I'm translating on Zanata, but there in sources, some strings was taget as fuzzy, but on Zanata not (or even strings are not present). So corrected/added here, on git.

### DIFF
--- a/po/cs.po
+++ b/po/cs.po
@@ -978,9 +978,8 @@ msgid "Random Number Generator"
 msgstr "Generátor náhodných čísel"
 
 #: ../virtManager/addhardware.py:818
-#, fuzzy
 msgid "VM Sockets"
-msgstr "Pa_tic:"
+msgstr "Patic virt. stroje"
 
 #: ../virtManager/addhardware.py:822 ../virtManager/details.py:2906
 #, python-format
@@ -2525,9 +2524,8 @@ msgid "Physical %s Device"
 msgstr "Fyzické %s zařízení"
 
 #: ../virtManager/details.py:2981
-#, fuzzy
 msgid "Cannot remove last video device while Graphics/Display is attached."
-msgstr "Není možné odebrat zařízení když je grafika/konzole připojená."
+msgstr "Není možné odebrat poslední zařízení když je grafika/konzole připojená."
 
 #: ../virtManager/details.py:3010 ../virtManager/details.py:3017
 #: ../virtManager/details.py:3023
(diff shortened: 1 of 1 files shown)


crobinso:
Thanks for the patch. However if they are correct in zanata, the next time I pull translations from zanata they should be updated correctly I think? Also one of these chunks is dangerous:


#: ../virtinst/cli.py:1324
#, fuzzy, python-format
#, python-format
msgid "Unknown %s options: %s"
msgstr "Neznámé volby: %s"

Notice the translation only has one %s in it, which would break the code in this case

p-bo:
Corrected, really sorry for that - did in hurry, when realized, that strings correctly translated on Zanata (where they looks OK), are tagged as fuzzy (thus unused) in source code in git :(
There on git, there were 11 strings marked as fuzzy, but when I went back to Zanata to fix them, they even wasn't here in this form (they are changed a bit on git). So when I cannot fix them on Zanata (git and Zanata are out of sync).
It is unfortunate, that I did 100% complete translation on Zanata before 2.1 release taging, but when I'm looking to 2.1 release archive, there is about 20 strings untranslated (fuzzy or not translated at all) - for future releases, how to achieve full translation before release please, when git is always step before Zanata?

crobinso:
Thank you for your efforts translating virt-manager! The only way we can get 100% coverage for a release is if we do a string freeze for a week or two before a release. Unfortunately with the way virt-manager development process works I don't think that's likely to happen, releases aren't on a regular schedule and are often done on short nitce. I can try to set up more regular automated pushing of strings to zanata which may help but it still won't be perfect if strings change right before release, which tends to happen with last minute bug fixing.

I've pushed the latest virt-manager.pot up to zanata now, does that enable you to correct the strings on the zanata side, or is there still an issue that requires direct patching? Generally I don't like to apply translation patches to the code directly if the zanata option works

p-bo:
> Thank you for your efforts translating virt-manager!
Glad to help, but only want to see work of us, translators, published to users…

> I can try to set up more regular automated pushing of strings to zanata which may help
It would be very helpful, please try to setup it.

> I've pushed the latest virt-manager.pot up to zanata now, does that enable you to correct the strings on the zanata side
Yes, it seems helped - but can we do import to git now to see, if there will remain some problems please?

crobinso:
Okay I've set up nightly push to zanata but only if there's new commits since the last push. And I've sync'd translations to master now. Please let me know if there's further issues

p-bo:
Looks good - no fuzzy strings, all translated (or left untraslatted as intended) and the best, it will be synchronized regularly - good work and many thanks for it!

PR #46, opened 2019-07-31
Add support for installing pre & post installable drivers

fidencio:
Windows supports installing pre & post installable drivers. Pre installable drivers are useful in order to have the guest disk using virtio instead of sata. Post installable drivers are useful in order to have spice-guest-tools, which brings all virtio drivers plus the QXL driver.

In order to have this done, I've also changed the unattended and installer* classes so those would be able to deal with multiple install scripts, which is exactly the Windows case.

### DIFF
--- a/virt-install
+++ b/virt-install
@@ -470,6 +470,12 @@ def set_cli_defaults(options, guest):
     if guest.os.is_container():
         return
 
+    if (options.unattended and
+        guest.osinfo.is_windows() and
+        guest.osinfo.supports_unattended_drivers(guest.os.arch)):
+        guest.add_extra_drivers(
+                guest.osinfo.get_pre_installable_devices(guest.os.arch))
+
     res = guest.osinfo.get_recommended_resources()
     storage = res.get_recommended_storage(guest.os.arch)
     ram = res.get_recommended_ram(guest.os.arch)
--- a/virtinst/guest.py
+++ b/virtinst/guest.py
@@ -487,13 +487,13 @@ class Guest(XMLBuilder):
         return False
 
     def supports_virtionet(self):
-        return self._supports_virtio(self.osinfo.supports_virtionet())
+        return self._supports_virtio(self.osinfo.supports_virtionet(self._extra_drivers))
     def supports_virtiodisk(self):
-        return self._supports_virtio(self.osinfo.supports_virtiodisk())
+        return self._supports_virtio(self.osinfo.supports_virtiodisk(self._extra_drivers))
     def supports_virtioscsi(self):
-        return self._supports_virtio(self.osinfo.supports_virtioscsi())
+        return self._supports_virtio(self.osinfo.supports_virtioscsi(self._extra_drivers))
     def _supports_virtioserial(self):
-        return self._supports_virtio(self.osinfo.supports_virtioserial())
+        return self._supports_virtio(self.osinfo.supports_virtioserial(self._extra_drivers))
 
 
     ###############################
@@ -838,7 +838,7 @@ class Guest(XMLBuilder):
             self.add_device(dev)
 
         # s390x guests need VirtIO input devices
-        if self.os.is_s390x() and self.osinfo.supports_virtioinput():
+        if self.os.is_s390x() and self.osinfo.supports_virtioinput(self._extra_drivers):
             dev = DeviceInput(self.conn)
             dev.type = "tablet"
             dev.bus = "virtio"
@@ -945,7 +945,7 @@ class Guest(XMLBuilder):
             return
 
         if (self.conn.is_qemu() and
-            self.osinfo.supports_virtiorng() and
+            self.osinfo.supports_virtiorng(self._extra_drivers) and
             self.conn.support.conn_rng_urandom()):
             dev = DeviceRng(self.conn)
             dev.type = "random"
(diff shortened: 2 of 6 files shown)


crobinso:
I pushed the first two patches

virtinst/install/installer.py
```diff
@@ -230,9 +234,28 @@ def _prepare_unattended_data(self, guest, scripts):
...
+        drivers_func = [guest.osinfo.get_pre_installable_drivers_location,
+                        guest.osinfo.get_post_installable_drivers_location]
+
+        for func in drivers_func:
+            drivers_location = func(guest.os.arch)
+            drivers = unattended.download_drivers(drivers_location,
+                    InstallerTreeMedia.make_scratchdir(guest))
+            if drivers:
+                # The cdrom with the drivers must be kept accross reboots,
+                # otherwise post-install drivers won't be installed
```
crobinso: Typo here and below 'accross'

virtinst/install/unattended.py
```diff
@@ -10,6 +10,7 @@
 import locale
 import os
 import pwd
+import requests
```
crobinso: The placement here triggers a pylint warning

virtinst/install/unattended.py
```diff
@@ -353,3 +354,17 @@ def _get_installation_source(os_media):
         script.set_config(config)
         scripts.append(script)
     return scripts
+
+
+def download_drivers(locations, tempdir):
```
crobinso: Using urlfetcher here will give us meter support, and http mocking. grep for DirectFetcher init and usage in installtreemedia.py, it can be used for downloading full URL paths by passing fullurl to acquireFile callback. You can use that to implement this function instead

virt-install
```diff
@@ -466,6 +466,12 @@ def set_cli_defaults(options, guest):
     if guest.os.is_container():
         return
 
+    if (options.unattended and
+        guest.osinfo.is_windows() and
+        guest.osinfo.get_pre_installable_drivers(guest.os.arch) and
+        guest.osinfo.get_post_installable_drivers(guest.os.arch)):
```
crobinso: It's a bit weird that we are just using these functions as boolean checks. Maybe make these private, but add a more descriptive function like osinfo.supports_unattended_drivers() that hides this logic

virtinst/guest.py
```diff
@@ -235,6 +235,7 @@ def __init__(self, *args, **kwargs):
         self.__osinfo = None
         self._capsinfo = None
         self._domcaps = None
+        self._use_virtio_win = False
```
crobinso: The way this is implemented I think it would be better to name it after what it's specifically doing, so something like force_supports_virtio. exposing that via the --install option too would be a nice addition for users who are manually installing virtio-win drivers. something for later though

virt-install
crobinso: Also an --unattended option to opt out of this behavior would be useful. Could be a follow up change though

crobinso:
I pushed the first five patches, with a bit of tests/osdict.py testing to use those functions for now.

I figured the issue you reported with file writing might have to do with a missing flush but it's weird I couldn't reproduce, are you using a non-standard filesystem setup? I'm using default ext4 on Fedora 30

The major issue with this is that the generated iso file is never removed from the guest config or removed from the FS. We don't follow this pattern for any other default generated or downloaded media so this is new territory.

Is it necessary for the unattended iso to be the source of postinstall drivers? Can the windows unattended process not instead stuff the spice tools installer exe inside the VM filesystem and have first boot access that instead? I realize that's likely more work but I think it's worth exploring if that's a simple change.

To make this work on virt-* side, we would either need to generate permanent media, meaning we upload it to the default pool instead of the scratch pool, and give it something like a stable name. But this is still annoying because the media really shouldn't need to exist forever, just a single extra boot. If we want to treat the media as transient and remove it after the second boot, we need to turn the install process into 3 phases will take work in virt-install and virt-manager and in the guts of installer.py which I would really like to avoid.

crobinso:
Also a smaller thing: the downloaded driver files are not added to _tmpfiles so they aren't cleaned up. I think maybe you should make download_drivers into download_driver, and do the iterating in installer, and add files as they are fetched to _tmpfiles

fidencio:
> I pushed the first five patches, with a bit of tests/osdict.py testing to use those functions for now.
> 
> I figured the issue you reported with file writing might have to do with a missing flush but it's weird I couldn't reproduce, are you using a non-standard filesystem setup? I'm using default ext4 on Fedora 30

I'm also using default ext4 on Fedora 30, but on an NVMe disk. Maybe that could cause some difference?

> 
> The major issue with this is that the generated iso file is never removed from the guest config or removed from the FS. We don't follow this pattern for any other default generated or downloaded media so this is new territory.
> 
> Is it necessary for the unattended iso to be the source of postinstall drivers? Can the windows unattended process not instead stuff the spice tools installer exe inside the VM filesystem and have first boot access that instead? I realize that's likely more work but I think it's worth exploring if that's a simple change.
> 

I'll have to dig into this and check whether we could copy something to, let's say, `C:\tmp\osinfo` and install it from there.

> To make this work on virt-* side, we would either need to generate permanent media, meaning we upload it to the default pool instead of the scratch pool, and give it something like a stable name. But this is still annoying because the media really shouldn't need to exist forever, just a single extra boot. If we want to treat the media as transient and remove it after the second boot, we need to turn the install process into 3 phases will take work in virt-install and virt-manager and in the guts of installer.py which I would really like to avoid.

libosinfo exposes the concept of number of reboots needed to finish an installation. GNOME Boxes takes advantage of that in order to decide what to do/when clean up the files.

crobinso:
Thanks! I pushed this series now, with a follow up patch to add the download drivers to the _tmpfiles list so they are cleaned up.

PR #51, opened 2019-09-06
osdict: Always return the most generic tree

fidencio:
Some OSes, as Fedora, have variants (which we rely to be standardised on
osinfo-db side), which we can use to return the most generic tree
possible, in case no profile is specified, in order to avoid failing to
install a "Workstation" system because a "Server" variant tree was used.

https://bugzilla.redhat.com/show_bug.cgi?id=1749865

NOTE: We **may** want to use the appropriate tree in the future as in using a Workstation tree for desktop installations or a Server tree for jeos installations. However, for now, let's not try to solve this issue as the "Everything" tree is a good enough fallback.

### DIFF
--- a/tests/osdict.py
+++ b/tests/osdict.py
@@ -52,10 +52,14 @@ class TestOSDB(unittest.TestCase):
 
     def test_tree_url(self):
         f26 = OSDB.lookup_os("fedora26")
+        f29 = OSDB.lookup_os("fedora29")
         winxp = OSDB.lookup_os("winxp")
 
         # Valid tree URL
         assert "fedoraproject.org" in f26.get_location("x86_64")
+
+        # Most generic tree URL
+        assert "Everything" in f29.get_location("x86_64")
 
         # Has tree URLs, but none for arch
         try:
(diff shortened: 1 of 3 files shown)


crobinso:
Thanks for the patch. That amount of indentation says to me we should move that logic to its own function, maybe just a private function on the OS object. I would also like to see test suite coverage. You can check the uncovered lines with `./setup.py test --coverage` then `coverage annotate -d covdir` and look at the osdict.py file in covdir. Simplest way to test it is probably via tests/osdict.py

fidencio:
> Thanks for the patch. That amount of indentation says to me we should move that logic to its own function, maybe just a private function on the OS object.

I'm not exactly sure if by "the OS object" you mean the _OsVariant one or something else. In any case, it's easy to change in case I got it wrong.

> I would also like to see test suite coverage. You can check the uncovered lines with `./setup.py test --coverage` then `coverage annotate -d covdir` and look at the osdict.py file in covdir. Simplest way to test it is probably via tests/osdict.py

Tests have been added, together with the change to also get a specific tree based on the profile used.

crobinso:
Thanks, pushed now

PR #55, opened 2019-10-02
Support "bochs" display

fidencio:
This series **tries** to add support to "bochs" display. Emphasis on tries as the code path where it'd be used is never ever triggered, as expressed in the commit message.

"bochs" display has also been added to virt-manager "Add Hardware" UI.

Note: In a conversation with Cole, we agreed that virt-install should revisit the video defaults in the near future in order to be aligned with Gerd's recommendations, present as part of this blog post https://www.kraxel.org/blog/2019/09/display-devices-in-qemu/ ... However, that's a different PR material.

### DIFF
--- a/virtinst/devices/video.py
+++ b/virtinst/devices/video.py
@@ -39,6 +39,9 @@ class DeviceVideo(Device):
             if guest.has_gl():
                 return "virtio"
             return "qxl"
+        if (guest.is_uefi() and
+            guest.lookup_domcaps().supports_video_bochs()):
+            return "bochs"
         if guest.conn.is_qemu():
             return "qxl"
         return "vga"
--- a/virtManager/addhardware.py
+++ b/virtManager/addhardware.py
@@ -663,7 +663,7 @@ class vmmAddHardware(vmmGObjectUI):
         if guest.conn.is_xen():
             return ["xen", "vga"]
         if guest.conn.is_qemu() or guest.conn.is_test():
-            return ["vga", "qxl", "virtio"]
+            return ["vga", "bochs", "qxl", "virtio"]
         return []
 
     @staticmethod
--- a/virtinst/domcapabilities.py
+++ b/virtinst/domcapabilities.py
@@ -95,6 +95,7 @@ class _Devices(_CapsBlock):
     XML_NAME = "devices"
     hostdev = XMLChildProperty(_make_capsblock("hostdev"), is_single=True)
     disk = XMLChildProperty(_make_capsblock("disk"), is_single=True)
+    video = XMLChildProperty(_make_capsblock("video"), is_single=True)
 
 
 class _Features(_CapsBlock):
--- a/virtinst/domcapabilities.py
+++ b/virtinst/domcapabilities.py
@@ -325,6 +325,14 @@ class DomainCapabilities(XMLBuilder):
         """
         return bool(self.features.sev.supported)
 
+    def supports_video_bochs(self):
+        """
+        Returns False if either libvirt or qemu do not have support to bochs
+        video type.
+        """
+        models = self.devices.video.get_enum("modelType").get_values()
+        return bool("bochs" in models)
+
     XML_NAME = "domainCapabilities"
     os = XMLChildProperty(_OS, is_single=True)
     cpu = XMLChildProperty(_CPU, is_single=True)


crobinso:
The first patch may cause us to select bochs when qemu or libvirt don't support it, meaning the VM fails to start. So it needs to be dependent on a domcaps check, hopefully that's already available in libvirt. We can get away with this in the other cases like because they have either been around for a very long time (qxl, vga), or we can assume they are supported for a fairly new arch (virtio+aarch64).

Patch #3, I'm trying to move away from that style of stuff, it adds complexity for not much gain IMO, makes testing difficult. I think if the new VM and addhardware default selection are correct then that's good enough

fidencio:
> The first patch may cause us to select bochs when qemu or libvirt don't support it, meaning the VM fails to start. So it needs to be dependent on a domcaps check, hopefully that's already available in libvirt. We can get away with this in the other cases like because they have either been around for a very long time (qxl, vga), or we can assume they are supported for a fairly new arch (virtio+aarch64).
> 

I'm not totally sure what we should check for in the domain capabilities. Are you suggesting we should add a new domain capability to return which are the video types supported?

The way we've discussed we would deal with bochs displays is basically fallbacking to it whenever a *guest* supports virtio (and now I see the patch may assume Windows would support it ...).

> Patch #3, I'm trying to move away from that style of stuff, it adds complexity for not much gain IMO, makes testing difficult. I think if the new VM and addhardware default selection are correct then that's good enough

Hmm. Fair enough.

crobinso:
> > The first patch may cause us to select bochs when qemu or libvirt don't support it, meaning the VM fails to start. So it needs to be dependent on a domcaps check, hopefully that's already available in libvirt. We can get away with this in the other cases like because they have either been around for a very long time (qxl, vga), or we can assume they are supported for a fairly new arch (virtio+aarch64).
> 
> I'm not totally sure what we should check for in the domain capabilities. Are you suggesting we should add a new domain capability to return which are the video types supported?
> 

Yes, but it's already there it seems, check `sudo virsh domcapabilities | grep qxl`, but I don't see bochs listed so maybe something is wrong. Since bochs is only valid for libvirt 5.6.0 + qemu-something, if we try to set bochs model on something older (like an existing RHEL8 system), the VM will completely fail to start. So we need domcapabilities to tell us if it's even a valid value before we try to use it.

This is really only important for setting defaults, not whether to list it in the UI IMO

> The way we've discussed we would deal with bochs displays is basically fallbacking to it whenever a _guest_ supports virtio (and now I see the patch may assume Windows would support it ...).

Since bochs is just a subset of 'vga' which has broad guest support, that's fine, but we can only do that when qemu + libvirt are new enough to not choke on the value

fidencio:
> > > The first patch may cause us to select bochs when qemu or libvirt don't support it, meaning the VM fails to start. So it needs to be dependent on a domcaps check, hopefully that's already available in libvirt. We can get away with this in the other cases like because they have either been around for a very long time (qxl, vga), or we can assume they are supported for a fairly new arch (virtio+aarch64).
> > 
> > 
> > I'm not totally sure what we should check for in the domain capabilities. Are you suggesting we should add a new domain capability to return which are the video types supported?
> 
> Yes, but it's already there it seems, check `sudo virsh domcapabilities | grep qxl`, but I don't see bochs listed so maybe something is wrong. Since bochs is only valid for libvirt 5.6.0 + qemu-something, if we try to set bochs model on something older (like an existing RHEL8 system), the VM will completely fail to start. So we need domcapabilities to tell us if it's even a valid value before we try to use it.

Aha, I missed that one. I'll re-work the patches and submit a new version soon.

> 
> This is really only important for setting defaults, not whether to list it in the UI IMO

Okay!

> 
> > The way we've discussed we would deal with bochs displays is basically fallbacking to it whenever a _guest_ supports virtio (and now I see the patch may assume Windows would support it ...).
> 
> Since bochs is just a subset of 'vga' which has broad guest support, that's fine, but we can only do that when qemu + libvirt are new enough to not choke on the value

Makes total sense.

Thanks for the review.

fidencio:
Cole,

I've re-worked the series by:
- Get video information from domain capabilities;
- Add a `supports_bochs_display` to both domain capabilities and guest;
- Checking whether bochs display is supported before setting it;

About the 3rd patch from the original series, it's been dropped.

crobinso:
Did you push the changes? I'm still seeing the old commits

Also I checked and indeed libvirt domcaps is missing code to add bochs to the video list. See virQEMUCapsFillDomainDeviceVideoCaps, it needs extending. It would also be nice to adjust the video model validation to instead use virDomainCapsDeviceDefValidate so this type of issue doesn't slip in again

fidencio:
Now I did. :-)

virtinst/devices/video.py
```diff
@@ -40,6 +40,8 @@ def default_model(guest):
         if guest.os.is_hvm():
             if guest.conn.is_qemu():
                 return "qxl"
+            if guest.supports_bochs_display():
+                return "bochs"
```
crobinso: In master, I moved the os.is_hvm() check to the top of the function, so this will collide a bit. But only because I think this should be tweaked a bit anyways

* Like you say, this won't ever be called. Your check should be moved before the is_qemu check. That qxl piece is meant to be a catchall for when the more specific cases above don't hit, but then not use qxl on xen.
* Rename supports_bochs_display to supports_video_bochs which will fit nicer if we add more of these helpers for video models in the future.
* The is_windows check should be hidden in a function like osinfo.supports_video_bochs().
* But I think guest.supports_bochs_display() should be dropped and instead opencode the logic here, so it's easier to follow the logic in a central place. The set_capabilities_defaults bit can be removed, it's not required. So here we would have `guest.is_uefi() and guest.osinfo.supports_video_bochs() and guest.lookup_domcaps().supports_bochs_display()`

fidencio: > In master, I moved the os.is_hvm() check to the top of the function, so this will collide a bit. But only because I think this should be tweaked a bit anyways
> 
> * Like you say, this won't ever be called. Your check should be moved before the is_qemu check. That qxl piece is meant to be a catchall for when the more specific cases above don't hit, but then not use qxl on xen.

Right.

> * Rename supports_bochs_display to supports_video_bochs which will fit nicer if we add more of these helpers for video models in the future.

Right.

> * The is_windows check should be hidden in a function like osinfo.supports_video_bochs().

I don't agree, Cole.
The **only** osinfo specific part is whether the guest is Windows or not. So,a `osinfo.supports_video_bochs()` would basically be `return not self.is_windows()` as the whole rest of the logic is part of the guest.

> * But I think guest.supports_bochs_display() should be dropped and instead opencode the logic here, so it's easier to follow the logic in a central place. The set_capabilities_defaults bit can be removed, it's not required. So here we would have `guest.is_uefi() and guest.osinfo.supports_video_bochs() and guest.lookup_domcaps().supports_bochs_display()`

Right!

crobinso: Re: the osinfo bit, the point is to just hide the osinfo specific logic inside a helper function there. Maybe it will grow in the future or need to be excluded for some certain distro some day, who knows. It's a minor distinction though so I'll leave it to your discretion

Though reading gerd's blog post I didn't find any mention of bochs + !windows, where does that bit of data come from?

fidencio: > Re: the osinfo bit, the point is to just hide the osinfo specific logic inside a helper function there. Maybe it will grow in the future or need to be excluded for some certain distro some day, who knows. It's a minor distinction though so I'll leave it to your discretion

I've adapted the patches according to your comments in the morning.
Yes, your point about expanding it in the future makes a lot of sense.

> 
> Though reading gerd's blog post I didn't find any mention of bochs + !windows, where does that bit of data come from?

So, Gerd never said explicitly it does not work on Windows. I've assumed that when I didn't see a Windows driver listed (as it's for QXL). But I also don't see one for VGA. :-/

I'll update the patches dropping that part.

crobinso:
I dropped the bit from the commit message about the code path not being triggered, and pushed. Thanks!

PR #58, opened 2019-10-18
Add "user-login" option for unattended installation

fidencio:
This series adds the possibility to pass "user-login" option when doing an unattended installation. Together with this change, we have to restrict the names used in order to avoid user reserved names (as root, administrator and a few other forbidden names for Windows).

This seems to be a reasonable solution for the case when an admin decides to run an unattended installation using either sudo or root, and the user set for the VM would be "root" ... breaking then the installation process for the most part of the distros (if not all).

### DIFF
--- a/man/virt-install.pod
+++ b/man/virt-install.pod
@@ -628,12 +628,18 @@ password-file.
 Note that only the first line of the file will be considered, including
 any whitespace characters and excluding new-line.
 
+=item B<user-login=>
+
+The user login name to be used in th VM. virt-install will default to your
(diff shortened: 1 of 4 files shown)


crobinso:
Thanks, I pushed the first patch. I don't want the second patch to be part of virt-install. If we want to explicitly reject 'root' which could be a common user mistake, that's one thing, but tracking the windows blacklist is not appealing to me. Either libosinfo/osinfo-db can team up to deal with that in some way, or we just ignore it

PR #60, opened 2019-10-24
fix resizing of spice clients with guest resizing enabled

FlorianLudwig:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1414718


crobinso:
Thanks! I dropped the commented out line from your patch and pushed it now

PR #62, opened 2019-11-08
installertreemedia: Respect --os-variant

fidencio:
**Problem**: When calling virt-install, explicitly passing `--os-variant`, depending on the `--location` used virt-install will consider the distro a "Generic" one and will **not** respect the passed `--os-variant` argument, resulting on a broken kernel command line (at least).

**Reproducer**:
```
./virt-install -r 4096 -n centos8-foobar  -w bridge=virbr0 --disk pool=default,size=10   --os-variant=rhel7.7 --location https://mirrors.mit.edu/cen
tos/8-stream/BaseOS/x86_64/os/ --extra-args='console=tty0 console=ttyS0,57600' --debug --dry-run
```
When using the reproducer above, you'll notice that inst.repo=$location_url will **not** be set.

**Notes**: Setting the os-variant seems to be quite (if not too much) tied to the guest / installer codes, which is not so easy to understand. Maybe there's a way simpler way to solve the issue and I'm more than happy on just dropping the patch in favour of a cleaner approach. However, this patch (which, IMHO, looks a little bit too hack-ish) was the best I could come up with.

### DIFF
--- a/virt-install
+++ b/virt-install
@@ -531,13 +531,16 @@ def set_explicit_guest_options(options, guest):
 
 def installer_detect_distro(guest, installer, osdata):
     try:
+        # OS name has to be set firstly whenever --os-variant is passed,
+        # otherwise it won't be respected when the installer creates the
+        # Distro Store.
+        if osdata.name:
+            guest.set_os_name(osdata.name)
+
         # This also validates the install location
         autodistro = installer.detect_distro(guest)
-        name = osdata.name
-        if osdata.is_auto:
-            name = autodistro
-        if name:
-            guest.set_os_name(name)
+        if osdata.is_auto and autodistro:
+            guest.set_os_name(autodistro)
     except ValueError as e:
         fail(_("Error validating install location: %s") % str(e))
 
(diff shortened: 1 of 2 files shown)


larsks:
This seems to work for me. With this PR, I can successfully boot a centos 8 installer using `--location` without needing to pass `inst.repo` via `--extra-args`.

virt-install
```diff
@@ -531,13 +531,19 @@ def set_explicit_guest_options(options, guest):
...
     try:
+        name = None
+
+        # OS name has to be set firstly whenever --os-variant is passed,
+        # otherwise it won't be respected when the installer creates the
+        # Distro Store.
+        if osdata.name and not osdata.is_auto:
+            name = osdata.name
+            guest.set_os_name(name)
+
```
crobinso: The changes in this file can be their own patch, they are a good to do independently I think

The name variable can be dropped.
The is_auto check here can be dropped too, `osdata.name` will be None if osdata.is_auto is True

virtinst/install/installertreemedia.py
```diff
@@ -180,6 +180,9 @@ def _get_cached_data(self, guest, fetcher):
         has_location_kernel = bool(
                 self._location_kernel and self._location_initrd)
 
+        if guest.osinfo.name != "generic":
+            os_variant = guest.osinfo.name
+
```
crobinso: You can use guest.osinfo.is_generic() here

crobinso:
Thanks, pushed now

PR #63, opened 2019-11-11
ADD  "Remove Local Path" menu,so that users can remove the local image if they want

skycastlelily2:
atm, we only have "Remove Hardware" menu which will only remove the device from xml file,
and virt-manager won't remove that local image when delete the VM,which will produce some
orphan images.We should add a "Remove Local Path" menu so that users can ask virt-manager
to remove the device and local path if they think they won't use that device anymore.

### DIFF
--- a/virtManager/delete.py
+++ b/virtManager/delete.py
@@ -31,6 +31,8 @@ STORAGE_ROW_TOOLTIP = 7
 
 
 class vmmDeleteDialog(vmmGObjectUI):
+
+    disk = None
     @classmethod
     def show_instance(cls, parentobj, vm):
         try:
@@ -75,6 +77,9 @@ class vmmDeleteDialog(vmmGObjectUI):
         self._set_vm(None)
         return 1
 
+    def set_disk(self, disk):
+        self.disk = disk
+
     def _cleanup(self):
         pass
 
@@ -93,9 +98,14 @@ class vmmDeleteDialog(vmmGObjectUI):
         self.vm = newvm
 
     def _reset_state(self):
-        # Set VM name in title'
+        # Set VM name or disk.target in title'
+        if self.disk:
+            text = self.disk.target
+        else:
+            text = self.vm.get_name()
+
         title_str = ("<span size='large' color='white'>%s '%s'</span>" %
-                     (_("Delete"), xmlutil.xml_escape(self.vm.get_name())))
+                     (_("Delete"), xmlutil.xml_escape(text)))
         self.widget("header-label").set_markup(title_str)
 
         self.topwin.resize(1, 1)
@@ -109,8 +119,10 @@ class vmmDeleteDialog(vmmGObjectUI):
         # Enable storage removal by default
         self.widget("delete-remove-storage").set_active(True)
         self.widget("delete-remove-storage").toggled()
-
-        diskdatas = _build_diskdata_for_vm(self.vm)
+        if self.disk:
+            diskdatas =[_DiskData.from_disk(self.disk),]
+        else:
+            diskdatas = _build_diskdata_for_vm(self.vm)
         _populate_storage_list(self.widget("delete-storage-list"),
                                self.vm, self.vm.conn, diskdatas)
 
@@ -172,10 +184,15 @@ class vmmDeleteDialog(vmmGObjectUI):
 
         self.set_finish_cursor()
 
-        title = _("Deleting virtual machine '%s'") % self.vm.get_name()
-        text = title
-        if devs:
-            text = title + _(" and selected storage (this may take a while)")
+        if self.disk:
+            title = _("Deleting the selected storage")
+            text = _('%s') % self.disk.target
+        else:
+            title = _("Deleting virtual machine '%s'") % self.vm.get_name()
+            text = title
+            if devs:
+                text = title + _(" and selected storage (this may take a while)")
+
 
         progWin = vmmAsyncJob(self._async_delete, [self.vm, devs],
                               self._delete_finished_cb, [],
@@ -195,18 +212,20 @@ class vmmDeleteDialog(vmmGObjectUI):
 
             conn = vm.conn.get_backend()
             meter = asyncjob.get_meter()
+            if not paths and self.disk:
+                vm.remove_device(self.disk)
 
             for path in paths:
                 try:
                     log.debug("Deleting path: %s", path)
                     meter.start(text=_("Deleting path '%s'") % path)
-                    self._async_delete_path(conn, path, meter)
+                    self._async_delete_dev(vm, conn, path, meter)
                 except Exception as e:
                     storage_errors.append((str(e),
                                           "".join(traceback.format_exc())))
                 meter.end(0)
 
-            if undefine:
+            if undefine and not self.disk:
                 log.debug("Removing VM '%s'", vm.get_name())
                 vm.delete()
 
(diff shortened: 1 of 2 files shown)


skycastlelily2:
There is another solution, add another checkbox below "Don't ask me again" by only modifying  the error.py and details.py,and set the checkbox's text to:Also want to remove the local image?.But the side affect is virt-manager will remove the local path without notice users after they click the "Don't ask me again" checkbox

crobinso:
Thanks for working on this! There's been an RFE around for a long time about it: https://bugzilla.redhat.com/show_bug.cgi?id=1058051

I think we should repurpose the delete.py wizard to handle this. We would need to change the labels in the UI some how, and conditionalize the code to not have anything to do with actually removing the VM. But the UI for selecting/deselecting whether you want to delete the storage, and determining whether we can _even_ delete the storage, is all the same code.

If you wanted to try and just copy the delete.py wizard to a different file and strip out the parts you don't need, just to get something working, that could be a start. If you get something functioning then we can discuss the next steps. I'm not sure if it should live as a separate file or just integrated all into delete.py

I don't think the 'dont ask again' bit is necessary. In this case we would just always default to not selecting storage for deletion, for back compat and safety, so if the users never want to remove storage then it's no more or less clicks to deal with

skycastlelily2:
Hi Cole,
   Thanks a lot for your feedback,I have tried to add a GtkCheckButton just above the "delete-remove-storage" one,now users  can choose to delete the disks( UI and the storage file) or the VM or both.Is that what's on your mind?If so,then I'm gonna to send a pull request right away.

crobinso:
Sounds reasonable, but I'll need to see the UI to be sure. You can `git push -f` over this branch and it will update the pull request, and I'll test it when I have time. Thanks!

skycastlelily:
Hi Cole,
   I have updated the pull  request,and thanks a lot for your time on reviewing it.

skycastlelily2:
Hi Cole, 
   I have updated the pull  request,and thanks a lot for your time on reviewing it^^

crobinso:
Okay I took a look. But I don't think that's the UI we want. The 'VM delete' case should stay exactly the same. The UI shouldn't change in that case.

What I meant, is to alter the delete.py code and UI to be able to used for just offering to delete a disk. So when you click 'remove' in the disk details page, you see a variant of the vmmDeleteDialog wizard. However the wizard won't touch the VM: it won't destroy or remove it. The UI will only show the existing storage list we have for the VM Delete wizard, and give you the option to remove the disk device and the host storage, or just the <disk> device from the XML

Maybe add a function like 'set_disk' to vmmDeleteDialog, which saves the disk we want to delete in `self.disk`. If that is present, use only that for populating the storage list (do `diskdatas = [_DiskData.from_disk(self.disk)]` and pass that to _populate_storage_list). In the `_finish` callbacks, if `self.disk` is available, it should ignore anything that touches `self.vm`. You may need to hide some UI pieces, or change some UI labels, to have the text make sense for this usecase. If you get something working, let me know here. I can help adjust it to better fit the existing code if necessary

skycastlelily2:
Hi Cole,
  I have updated the pull request,and thanks for your time again,and sorry for the noise
 (I forgot to git commit,before run git push -f)

crobinso:
It looks like your merged your patch based on old code into the new code, but this stripped out cleanup changes I added to delete.py to try and simplify your additions.

Notice the 'nitialization methods' and 'UI init helpers' are removed by your patch. That shouldn't happen. Nor should functions be deprivatized (_reset_state renamed to reset_state)

`populate_storage_list` shouldn't need to change its signature either. your changes should generate the `diskdata` for just the relevant disk, and pass it to `populate_storage_list`

skycastlelily2:
Hi Cole,
   Really sorry for having made such a big noise,and really thanks a lot for your patience and time on guiding me^^

virtManager/delete.py
```diff
@@ -31,6 +31,12 @@
 
 
 class vmmDeleteDialog(vmmGObjectUI):
+    disk = None
+
+    @classmethod
+    def set_disk(cls, disk):
+        cls.disk = disk
+
```
crobinso: You don't want to use a @classmethod here. That means that you are setting vmmDeleteDialog.disk, but you want to be setting self.disk. So this should be a regular method:
```python
def set_disk(self, disk):
    self.disk = disk
````

File: virtManager/details/details.py [hunk 2]
```diff
@@ -606,6 +606,7 @@ def init_menus(self):
 
         rmHW = Gtk.ImageMenuItem.new_with_label(_("_Remove Hardware"))
         rmHW.set_use_underline(True)
+        rmHW.topwin = self.topwin
```
crobinso: any setting `topwin = self.topwin` is a sign that something is wrong

virtManager/details/details.py
```diff
@@ -1110,7 +1111,10 @@ def add_hardware(self, src_ignore):
 
     def remove_xml_dev(self, src_ignore):
         devobj = self.get_hw_row()[HW_LIST_COL_DEVICE]
-        self.remove_device(devobj)
+        from ..delete import vmmDeleteDialog
+        vmmDeleteDialog.set_disk(devobj)
+        vmmDeleteDialog.show_instance(src_ignore, self.vm)
+
```
crobinso: The reason you had to set `topwin` elsewhere is because you are passing `src_ignore` here. You should be passing in `self`

This is going to attempt to launch vmmDeleteDialog for every device removal, not just disk removal. I think a good first patch would be to change this remove_xml_dev function. Have it check `devobj.DEVICE_TYPE == "disk"` and if try, calls a function self.remove_disk. For a first start it should just call self.remove_device(), so nothing changes. All other devices will continue to call self.remove_device(). That could be committed first.

remove_disk() function can then handle initializing vmmDeleteDialog in the right way, and launching it. I don't think it should use show_instance, that's complicating things too much. Just have it do:

```python
dialog = vmmDeleteDialog()
dialog.set_disk(disk)
dialog.show(self.topwin, self.vm)
```

skycastlelily2:
Hi Cole,
   Really thanks a lot for your guide,I feel I have gained many knowledge even though my PR has not been merged yet. By "a good first patch" you mean I should modify the code in different commits,right? So,I have sent two commits.Two commits in virt-manager,yay^^

crobinso:
I pushed the patches. Congratulations! I'll close this. However there's still some issues and more work to be done.

* Try this on a running VM, you will see the delete dialog shows a text warning that the VM is running. That should be hidden
* Ideally we would first time the device removal operation. If that fails, we dont remove the storage. This will be tricky to do. I think vmmDetails.remove_device function will need to be passed to the dialog, and that is called first before performing the storage deletion. You'll probably need to move the yes/no confirmation handling out of remove_device, and only use that for non-disk devices

Ideally the code would use a separate class from vmmDeleteDialog. Here's an example arrangement that would be nice.

```python
class _vmmDeleteBase:
    # This has all the code of vmmDeleteDialog
    def _get_dialog_title(self):
        raise NotImplementedError

    def _reset_state(self):
        text = self._get_dialog_title()
        title_str = ("<span size='large' color='white'>%s '%s'</span>" %         
                     (_("Delete"), xmlutil.xml_escape(text)))                    
        self.widget("header-label").set_markup(title_str)  
        # insert remaining code here

class vmmDeleteDialog(_vmmDeleteBase):
    def _get_dialog_title(self):
        return self.vm.get_name()

class vmmDeleteStorage(_vmmDeleteBase):
    def _get_dialog_title(self):
        return self.disk.target
```

The idea is that we try to hide all the `self.vm` vs `self.disk` differences into individual functions, that `_vmmDeleteBase` will act on. The reason for this, besides making the code easier to follow, is so that future changes make it harder to accidentally use `self.vm` or `self.disk` when they shouldn't be touching those values. So basically every `vmmDeleteDialog` piece that checks for `if self.disk` should be replaced by a function that will be overwritten by a subclass. That's something you could implement function by function with a different commit in each piece, so it could generate a lot of commits ;)

PR #65, opened 2019-11-20
unattended: Never use "root" as user-login

fidencio:
When running virt-install as root, user-login would be automatically set
to "root", causing an installation failure in the most part of the
distros (if not all of them).

In order to avoid such failures, let's raise a runtime error in case the
user-login used is "root".


crobinso:
Thanks, pushed now

PR #67, opened 2019-12-05
virtManager: disambiguize LXC

cpaelzer:
LXC can be many things and in particular when seeing just "LXC" in the
virt-manager UI one might think of either
- Libvirt-LXC => https://libvirt.org/drvlxc.html
- Linuxcontainers LXC => https://linuxcontainers.org

Especially with the "Linux Containers" that follows in brackets that is misleading. Lets get rid of this ambiguity and name the backend clearly to be the former and not the latter.

This change is in [Ubuntu since 2015](https://git.launchpad.net/ubuntu/+source/virt-manager/tree/debian/patches/mark-libvirt-lxc.patch?h=applied/ubuntu/focal-devel) so it might have been discussed (and denied?) in another context already. I feel like even I might have discussed it once, but too long ago to remember or find it.
Never the less I think this is helpful to the users in general so lets try (again?)

### DIFF
--- a/po/as.po
+++ b/po/as.po
@@ -1373,10 +1373,6 @@ msgstr "সংযোগ ডাইলগ আৰম্ভ কৰোতে ত্�
 msgid "user session"
 msgstr ""
 
-#: ../virtManager/createconn.py:119
-msgid "Linux Containers"
-msgstr ""
-
 #: ../virtManager/createconn.py:241
 msgid "A hostname is required for remote connections."
 msgstr "দূৰৱৰ্তী সংযোগসমূহৰ বাবে এটা হস্টনামৰ প্ৰয়োজন।"
--- a/po/bg.po
+++ b/po/bg.po
@@ -1299,10 +1299,6 @@ msgstr ""
 msgid "user session"
 msgstr ""
 
-#: ../virtManager/createconn.py:119
-msgid "Linux Containers"
-msgstr ""
-
 #: ../virtManager/createconn.py:241
 msgid "A hostname is required for remote connections."
 msgstr ""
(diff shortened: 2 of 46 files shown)


Antique:
I'm not so sure about this change. We don't use Libvirt-QEMU for QEMU/KVM as well even though there is also [https://libvirt.org/drvqemu.html](https://libvirt.org/drvqemu.html).

cpaelzer:
@Antique thanks for your reply, but I think that is a bad example. We have to consider what people usually think of when you say those words. When virt-manager displays "Qemu/KVM" people will assume https://libvirt.org/drvqemu.html because that is the first management layer that you'd hear/find searching or talking about it.

But for "LXC (Linux Containers)" that is not true. If you e.g. go to a search engine or mention those three words at a conference people will more likely related to https://linuxcontainers.org/ than https://libvirt.org/drvlxc.html. Because there are more common management layers qualifying for those three words - even the command `lxc` will be for LXC/LXD and not libvirt-lxc.
So really I think the dis-ambiguity would help a lot in this case.

crobinso:
I under Pavel's objection but personally I'm fine a with a change like this. 'lxc' in isolation is ambiguous, libvirt's lxc driver is completely distinct from cli lxc usage. In fact I'd go further and just change the label to 'Libvirt LXC', even listing 'linux containers' is quite ambiguous and is only likely to encourage people to select that option when they probably shouldn't.

cpaelzer:
@crobinso ok, I've added a patch that drops the following "Linux Containers" as well.

crobinso:
I squashed the patches, dropped the po/ changes because those will be dealt with next time we sync translations, and pushed the patch

PR #68, opened 2019-12-12
installerinject: Check for genisoimage and mkisofs before calling it

fidencio:
Let's check for both genisoimage and mkisofs as some distros will ship
mkisofs without linking it to genisoimage, as OpenSUSE Leap 15.1 does.

And, I know, genisoimage is a requirement of the project. However, it's
not packaged on OpenSUSE Leap 15.1 while mkisofs is present there.


crobinso:
Thanks, pushed now

PR #69, opened 2019-12-19
Handle the warning messages for disk device and non-disk device differently

skycastlelily2:
add a remove-non-disk function and move the yes/no confirmation handling there


skycastlelily2:
Hi Cole,
  Here is another commit,I'm gonna to send more in the following days,lots of commits sounds very yummy and I feel I will learn more during that process ,thanks a lot for your help and guide again:)

crobinso:
Thanks, pushed now

PR #76, opened 2020-01-22
Pass vmmDetails's remove_device function to vmmDeleteStorage

skycastlelily2:
the remove_device function in vmmDetails handles more things than vm.remove_device.
We should call it first before performing the storage deletion.

### DIFF
--- a/virtManager/delete.py
+++ b/virtManager/delete.py
@@ -305,6 +305,37 @@ class vmmDeleteDialog(_vmmDeleteBase):
 
 
 class vmmDeleteStorage(_vmmDeleteBase):
+    @staticmethod
+    def remove_devobj_internal(vm, err, devobj):
+        log.debug("Removing device: %s", devobj)
+
+        # Define the change
+        try:
+            vm.remove_device(devobj)
+        except Exception as e:
+            err.show_err(_("Error Removing Device: %s") % str(e))
+            return
+
+        # Try to hot remove
+        detach_err = ()
+        try:
+            if vm.is_active():
+                vm.detach_device(devobj)
+        except Exception as e:
+            log.debug("Device could not be hotUNplugged: %s", str(e))
+            detach_err = (str(e), "".join(traceback.format_exc()))
+
+        if not detach_err:
+            return True
+
+        err.show_err(
+            _("Device could not be removed from the running machine"),
+            details=(detach_err[0] + "\n\n" + detach_err[1]),
+            text2=_("This change will take effect after the next guest "
+                    "shutdown."),
+            buttons=Gtk.ButtonsType.OK,
+            dialog_type=Gtk.MessageType.INFO)
+
     def __init__(self, disk):
         _vmmDeleteBase.__init__(self)
         self.disk = disk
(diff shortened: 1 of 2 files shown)


virtManager/delete.py
```diff
@@ -73,8 +73,6 @@ def _delete_vm(self, vm):
 
     def _delete_disks(self, vm, paths, conn, meter):
         storage_errors = []
-        if not paths:
-            return
```
crobinso: Please submit this as a separate patch. You can subject it: 'delete: remove redundant code'

crobinso:
Hmm looking at this now I think I led you in the slightly wrong direction. I think we need to break out the vmmDetails.remove_device logic to mostly be a static function, otherwise the dialogs will be raised on top of the details window, and not the 'delete' window like we want.

First thing you will need to do is change vmmDetails.remove_device() to not call 'disable_apply'. Split the function so that vmmDetails remove_device now looks like:

```python
def _remove_device_internal(self, devobj):
   # previous devobj code except it has
   if not detach_err:
        return True
   # ...

def remove_device(self, devobj):
    success = self._remove_device_internal(devobj)
    if not success:
        self.disable_apply()
```

That can be one commit. Then the next commit can change _remove_device_internal to be:

```python
def _remove_device_internal(self, vm, err, devobj):
    # remove all usage of 'self' and use 'vm' and 'err' directly

def remove_device(self, devobj):
    success = self._remove_device_internal(self.vm, self.err, devobj)
    # ...
```

third commit, move _remove_devobj_internal to be a static function vmmDeleteStorage, call it remove_device. Use it in details.py

final commit, use that function in delete.py as necessary.

skycastlelily2:
Hi Cole,
  I have modified the code accordingly,thanks a lot for your guide again:)

virtManager/details/details.py
```diff
@@ -1967,6 +1965,11 @@ def remove_device(self, devobj):
             buttons=Gtk.ButtonsType.OK,
             dialog_type=Gtk.MessageType.INFO)
 
+    # Device removal
+    def remove_device(self, devobj):
+        success = self.remove_devobj_internal(devobj)
+        if success:
+            self.disable_apply()
```
crobinso: You removed the newline after the function. Please fix this in the next version. I suggest visually reviewing each diff before you submit things, the changes will stand out more

virtManager/details/details.py
```diff
@@ -1935,7 +1935,8 @@ def config_vsock_apply(self, devobj):
                                           kwargs, self.vm, self.err,
                                           devobj=devobj)
 
-    def remove_devobj_internal(self, vm, err, devobj):
+    @staticmethod
+    def remove_devobj_internal(vm, err, devobj):
```
crobinso: Actually I think this function should be moved to vmmDeleteStorage class, then the callback doesn't need to be passed into the init function, and vmmDetails can call that function directly.

Put staticmethod functions at the top of the class definition, above the `__init__` function

crobinso:
Thanks! I've pushed these now

PR #77, opened 2020-01-24
unattended: Add reg-login option to --unattended

fidencio:
As osinfo-db introduced the first usage of reg-login, let's also
add support for such option when using --unattended.

NOTE: This is the osinfo-db commit which adds the first usage of reg-login: https://gitlab.com/libosinfo/osinfo-db/merge_requests/87
The API has been there forever, but not being used.


crobinso:
Pushed, thanks!

PR #101, opened 2020-04-20
urldetect: fixup discovery of legacy ubuntu focal images

xnox:
In the future, virt-manager will need to switch to
cloud-images.ubuntu.com or maas.io images or live-server images, as
d-i builds are going away.

Signed-off-by: Dimitri John Ledkov <xnox@ubuntu.com>

### DIFF
--- a/virtinst/install/urldetect.py
+++ b/virtinst/install/urldetect.py
@@ -648,6 +648,8 @@ class _DebianDistro(_DistroTree):
         media_type = None
         if check_manifest("current/images/MANIFEST"):
             media_type = "url"
+        elif check_manifest("current/legacy-images/MANIFEST"):
+            media_type = "url"
         elif check_manifest("daily/MANIFEST"):
             media_type = "daily"
         elif cache.content_regex(".disk/info",
@@ -701,6 +703,9 @@ class _DebianDistro(_DistroTree):
 
     def _set_url_paths(self):
         url_prefix = "current/images"
+        if self._debname == "ubuntu" and self._os_variant >= 'ubuntu20.04':
+            url_prefix = "current/legacy-images"
+            log.warn("Using legacy d-i based installer, that has been deprecated and will be removed in the future. https://discourse.ubuntu.com/c/server")
         if self.cache.debian_media_type == "daily":
             url_prefix = "daily"
         elif self.cache.debian_media_type == "mounted_iso_url":


crobinso:
Thanks, can you provide an example URL to test with? I'd like to add it to the test_urls.py functional test suite to ensure we don't regress

xnox:
@crobinso 

http://us.archive.ubuntu.com/ubuntu/dists/focal/main/installer-amd64/
I think is the right format URL that a user would pass, but has a new structure further down inside it.

For comparison: http://us.archive.ubuntu.com/ubuntu/dists/bionic/main/installer-amd64/ is the URL that a user would pass, but has the old structure.

Both should find all the right pieces virt-install needs (like kernel & initrd), with focal using `/legacy-images/` somewhere down the line, whereas bionic is using `/images/` somewhere down the line.

XeS0r:
I was able to trigger an exception when _os_variant is None. Even if you set `--os-variant`.

```bash
virt-install \
--name ubuntu20-guest1 \
--memory 4096 \
--vcpus 2 \
--disk "size=4" \
--location "http://bistro.lnxne.boe/ubuntu/UBUNTU20.04/CD,kernel=boot/kernel.ubuntu,initrd=boot/initrd.ubuntu" \
--os-variant "ubuntu20.04" \
--network "network=default" \
--initrd-inject "/root/kvm-guests/ubuntu20-guest1/preseed.cfg" \
--extra-args "auto-install/enable=true"

```

```
> /usr/share/virt-manager/virtinst/install/urldetect.py(698)_set_url_paths()
-> if self._debname == "ubuntu" and self._os_variant >= 'ubuntu20.04':
(Pdb) w
  /usr/share/virt-manager/virt-install(994)main()
-> guest, installer = build_guest_instance(conn, options)
  /usr/share/virt-manager/virt-install(557)build_guest_instance()
-> installer_detect_distro(guest, installer, osdata)
  /usr/share/virt-manager/virt-install(525)installer_detect_distro()
-> autodistro = installer.detect_distro(guest)
  /usr/share/virt-manager/virtinst/install/installer.py(386)detect_distro()
-> ret = self._treemedia.detect_distro(guest)
  /usr/share/virt-manager/virtinst/install/installertreemedia.py(300)detect_distro()
-> cache = self._get_cached_data(guest, fetcher)
  /usr/share/virt-manager/virtinst/install/installertreemedia.py(185)_get_cached_data()
-> store = urldetect.getDistroStore(guest, fetcher,
  /usr/share/virt-manager/virtinst/install/urldetect.py(292)getDistroStore()
-> store = sclass(fetcher.location, arch, _type, cache)
  /usr/share/virt-manager/virtinst/install/urldetect.py(349)__init__()
-> self._set_manual_kernel_paths()
  /usr/share/virt-manager/virtinst/install/urldetect.py(669)_set_manual_kernel_paths()
-> self._set_url_paths()
> /usr/share/virt-manager/virtinst/install/urldetect.py(698)_set_url_paths()
-> if self._debname == "ubuntu" and self._os_variant >= 'ubuntu20.04':
(Pdb) p vars(self)
{'type': 'hvm', 'arch': 's390x', 'uri': 'http://bistro.lnxne.boe/ubuntu/UBUNTU20.04/CD', 'cache': <virtinst.install.urldetect._DistroCache object at 0x3ff9c0f9d00>, '_os_variant': None, '_kernel_paths': []}
```

synergiator:
please comment: is the following virt-install error (using --location URL) related to: a) this PR b) need to use another URL c) could use an ISO image instead for amd64  (as current desktop image is not working, same problem)
`
ValueError: Could not find an installable distribution at 'http://archive.ubuntu.com/ubuntu/dists/focal/main/installer-amd64/current/legacy-images/netboot/ubuntu-installer/amd64/'`

funkyfuture:
@synergiator, if i followed [this thread](https://bugs.launchpad.net/ubuntu-z-systems/+bug/1872941) correctly, yes it relates to this PR. maybe @xnox can clarify if needed. but i haven't read how long the legacy installer is supposed to be kept.

xnox:
> @synergiator, if i followed [this thread](https://bugs.launchpad.net/ubuntu-z-systems/+bug/1872941) correctly, yes it relates to this PR. maybe @xnox can clarify if needed. but i haven't read how long the legacy installer is supposed to be kept.

Normally ubuntu publishes two streams of d-i based images /$series/ and /$series-updates/. The first one corresponds to GA release, and the other one corresponds to point releases. Thus /focal/ will be there for the lifetime of focal itself. It is to be determined if any /focal-updates/ are to be published. At the moment there are no plans to publish /focal-updates. And /groovy/ currently has legacy-images published, but they are expected to be removed before the release.

Thus /legacy-images/ support for focal release is needed for at least next ten years.

xnox:
> ```shell
> http://bistro/ubuntu/UBUNTU20.04/CD
> ```



> I was able to trigger an exception when _os_variant is None. Even if you set `--os-variant`.
> 
> ```shell
> virt-install \
> --name ubuntu20-guest1 \
> --memory 4096 \
> --vcpus 2 \
> --disk "size=4" \
> --location "http://bistro/ubuntu/UBUNTU20.04/CD,kernel=boot/kernel.ubuntu,initrd=boot/initrd.ubuntu" \

I don't know what above means. Can you please elaborate what is located at bistro/ubuntu/UBUNTU20.04/CD ? I.e. "it's the same as this public URL" or like "it's the mounted public iso as seen on http://cdimage.ubuntu.com/something" ?

ptoscano:
@xnox it looks like @fidencio a slightly different version as #112; his version seems better to me:
- no hardcoded distribution
- no usage of a `None` variable, as reported by @XeS0r
- no useless warning about the legacy installer; using the new installer requires changes in virt-manager and possibly libosinfo, so it is not something the user can do anything about (which means that the warning is not actionable)
- includes the tests for this

Can you (and all the other commenters to this MR) try #112, please?

fidencio:
Hey @xnox, sorry, my bad as I didn't realise you already had opened a PR fixing that issue :-/

I've faced that while testing a completely different thing and took the bullet to also have that fixed while it was fresh. I didn't mean to step on your toes, in any possible way.

Please, if you (as Pino) also consider my solution simpler, let's get @crobinso's input and go for that.
If not, I'm also fine about dropping my patch as, again, I didn't mean to step on your toes by opening my PR.

Hope we can have any of the solutions merged. :-)

xnox:
Either PR is fine.

The warning about legacy installer, imho should be shown. The point of renaming the URLs is to warn that we will stop publishing these images, and that software has to be fixed and users will have to migrate to the new installer or preinstalled images, as it seem fit.

ptoscano:
> The warning about legacy installer, imho should be shown. The point of renaming the URLs is to warn that we will stop publishing these images, and that software has to be fixed and users will have to migrate to the new installer or preinstalled images, as it seem fit.

As I mentioned before, I disagree with this. Showing to the users a warning they cannot do anything about is not useful, and will only result in bug reports like "please fix or remove this warnings" (because that's the only thing an user can do).

There is already #106 for this, plus https://gitlab.com/libosinfo/osinfo-db/-/issues/54 for osinfo-db, and that's all is enough for users.

crobinso:
I pushed fabiano's patch, it came with tests so made my life simpler, and was more generalized. Sorry for not getting to this sooner though

PR #109, opened 2020-05-13
Fix Send Key translation in pt_BR

hugomg:
This fixes a small mistake in the pt_BR translation. Currently it is translating "key" with the same word used for cryptographic keys instead of the word for keyboard keys.

Enviar Chave -> Enviar Tecla


crobinso:
Thanks for the contribution!

PR #112, opened 2020-05-21
urldetect: Deal with ubuntu 20.04 legacy installer

fidencio:
Ubuntu 20.04 has a new installer, which is yet not supported by
virt-install / osinfo-db, and this made then switch their URLs
for the old installer to current/legacy-images/... instead of
current/images/...

Let's adapt URL detect so it can deal with thsi new "legacy"
style.


XeS0r:
I can confirm that this works fine with s390x installations of ubuntu-20.04.

crobinso:
I've pushed this now, thanks!

PR #113, opened 2020-05-21
Fix typo in virt-install man page

tonyjames:
The flag for hiding the KVM hypervisor signature has too many "=" signs.


crobinso:
Thanks for the contribution!

PR #133, opened 2020-07-24
Fix duplicate machine type entry in aarch64 when creating new VM

t-msn:
Hello

Just noticed that when creating new aarch64 machine, "virt-5.0" entry appeared twice.
This PR tries to fix this problem by avoiding the same alias name.

before patch:

after patch:


phrdina:
Looks good and fixes the issue. Thanks for the patch.

PR #134, opened 2020-07-24
CONTRIBUTING: Fix uitests test name

t-msn:
"python -m pytest --uitest" does not work as expected.


phrdina:
Both works for me but pytest --help mentions --uitests so this should be updated. Thanks for the patch.

PR #137, opened 2020-08-10
Add USB cdrom and ramfb video

jmglogow:
Both are needed to install Windows 10 Arm64 from ISO.

Just FYI, if you want to test this: libvirt currently doesn't create correct USB cdrom devices, which breaks the installer. More info is available in the pending merge request for it: https://gitlab.com/libvirt/libvirt/-/merge_requests/19


crobinso:
Thanks for the patch. Sorry about the CI failure

PR #163, opened 2020-10-07
Use InputsChannel instead of DisplayChannel when video is "none"

sjnewbury:
When Video is "none", SPICE never opens the DisplayChannel.
Currently the code just waits forever for the connection, this
means it's impossible to use SPICE input/audio when using
GPU-passthrough.

Instead of waiting for DisplayChannel, use InputsChannel instead.
This works, but doesn't provide the width and height so test for
the DisplayChannel rather than hitting an exception.

### DIFF
--- a/virtManager/details/viewers.py
+++ b/virtManager/details/viewers.py
@@ -605,6 +605,26 @@ class SpiceViewer(Viewer):
                 self._main_channel, "notify::agent-connected",
                 self._agent_connected_cb)
 
+        elif (type(channel) == SpiceClientGLib.InputsChannel and
+                not self._display):
+            # If there is no video output initialse display with inputs
+            # only. DisplayChannel doesn't exist so connection never occurs.
+            if not any([v for v in self._vm.get_xmlobj().devices.video if
+                    v.model == "none"]):
+                return	# pragma: no cover
+
+            channel_id = channel.get_property("channel-id")
+
+            if channel_id != 0:  # pragma: no cover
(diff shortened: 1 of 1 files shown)


crobinso:
Thanks for the patch. Can you clarify how this is actually used? You have graphical output going from the VM GPU to your monitor, but on the host you have the console window fullscreen so input/audio 'just works' without transferring control to the VM? Something like that?

sjnewbury:
> Thanks for the patch. Can you clarify how this is actually used? You have graphical output going from the VM GPU to your monitor, but on the host you have the console window fullscreen so input/audio 'just works' without transferring control to the VM? Something like that?

No need for the console window to be full screen.  WIth input focus the keyboard is relayed through ps2/virtio keyboard.  When pointer is grabbed it too is relayed.  Latency is pretty reasonable.  Leave pointer lock/unfocus works as you'd expect, exactly as usual.

crobinso:
Interesting, it's a clever idea for sure. But this is not suitable for virt-manager IMO. See [DESIGN.md](https://github.com/virt-manager/virt-manager/blob/master/DESIGN.md) for some background. Particularly

* We are aiming to not provide any more functionality than provided by other clients
* We would be deviating from other clients (virt-viewer, spicy, vinagre) in a very common code path, so we would be at greater risk for hitting unique bugs
* This is an advanced niche usecase
* This code can only be triggered via a very specific VM config which expands our test matrix.

So unfortunately this will need to live as a custom patch. Sorry.

PR #166, opened 2020-10-14
Add toggle menu to enable/disable SPICE clipboard sync

thebigb:
While it's a nice feature, it bothered me a bit that clipboard sync is automatically enabled (unless explicitly disabled in the XML), and there is no way to toggle it while the VM is running.

This pull request adds a toggle menu to the VM window (the setting is saved per VM), and adds a global setting of the default state.

screen-vmwindow

screen-prefs

I built this for myself, and figured I'd try to upstream this. I've seen `DESIGN.md`, so I suspect it's a bit out of scope, but I'll let you be the judge of that.

### DIFF
--- a/data/org.virt-manager.virt-manager.gschema.xml
+++ b/data/org.virt-manager.virt-manager.gschema.xml
@@ -31,6 +31,12 @@
       <summary>Autoconnect to the default VM console when the VM window is opened</summary>
       <description>Autoconnect to the default VM console when the VM window is opened. Users may want to turn this off if they prefer to use another viewer app for their VMs, and don't want virt-manager to interfere, but they still want to use virt-manager's details. -1 = global default, 0 = off, 1 = on</description>
     </key>
+
+    <key name="auto-clipboard" type="i">
+      <default>-1</default>
+      <summary>Enable SPICE clipboard sync by default in the console window</summary>
+      <description>Whether to enable SPICE clipboard sync by default while connected to the guest console.</description>
+    </key>
   </schema>
 
 
@@ -227,6 +233,12 @@
       <default>true</default>
       <summary>Enable SPICE Auto USB redirection in console window</summary>
       <description>Whether to enable SPICE Auto USB redirection while connected to the guest console.</description>
+    </key>
+
+    <key name="auto-clipboard" type="b">
+      <default>true</default>
+      <summary>Enable SPICE clipboard sync by default in the console window</summary>
+      <description>Whether to enable SPICE clipboard sync by default while connected to the guest console.</description>
     </key>
 
     <key name="autoconnect" type="b">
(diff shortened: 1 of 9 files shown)


crobinso:
Nice patch! I didn't review it deeply but it seems to hit all the bases. 

But like you predicted I am unsure about adding this to virt-manager. A global toggle in Preferences isn't too bad but I don't see much value in a per-VM override when there's the XML option to disable clipboard.

Can you explain your usecase a bit more? Do you change the default globally but added the per-VM toggle for completeness, or do you need this on a per-VM basis?

The other thing worth mentioning is that you can effectively disable spice clipboard usage by setting SPICE_MAX_CLIPBOARD=0 environment variable

thebigb:
The use-case is having secrets on the host, for example a password manager, and being able to copy a secret to only a single guest when you have multiple VMs running. So yes, the per-VM toggle is an important mechanic for this. There are workarounds like auto-type, but in my experience that isn't always as reliable with guests under high load or certain operating systems (looking at you Windows).

The XML option isn't practical because it can't be modified on the fly (besides it being inconvenient having to add XML every time you need it).

crobinso:
Okay so sounds like you default to clipboard disabled, but enable it per VM on demand when pasting a secret, and disable it right after. That's interesting, but not something I think fits with the rationale laid out in DESIGN.md, sorry. Thank you for submitting the patch though

summer-espresso:
I understand the point of view but I would say this is somewhat unfortunate because personally I think that by default guest systems should not be considered trusted (unlike https://github.com/virt-manager/virt-manager/issues/358#issuecomment-1031366985)

My use case (and that's why I use a VM) is that the contents of that VM are not considered trusted and I have several like that.
So I would like that by default there is no transmission of clipboard information except the few times I need it and for which I would manually activate the clipboard, like VirtualBox. This would save me from having to shut down and restart the VM just to have to copy and paste.

zoechi:
I think the ideal way would be like it's done in QubesOS where an explicit keyboard shortcut (Ctrl+Shift+C, Ctrl+Shift-V) is used to copy to and from the shared clipboard.

WayneSallee:
I think it's sad that this patch was not accepted.

Having clipboard enabled by default is a huge security risk. A program running in the guest could be constantly reading the clip board. 

On VirtualBox you can easily click to enable or disable clipboard per guest, on the fly. Not only that, but you can choose to only allow one direction or the other. And also can turn network on or off.

It's not at all unusual for people to use Virtual software to test programs that they don't trust. The virtual software should act as a firewall, and only open connections where the user request connections to be open.

So if you don't like this patch, then what about having a default settings of <clipboard copypaste="no"/> And people can change it in the default settings to enable clipboard. At least that would make it more secure, and allow people to know how to enable per guest. But it still would require a guest shut down (reboot won't work) for each change.

I like zoechi's comment of using a special key combination for copy and paste,

Wayne Sallee
Wayne@WayneSallee.com

PR #175, opened 2020-11-03
virtinst: Fix TOCTOU race in user scratch dir creation

martinpitt:
When starting many virt-install processes in parallel, some often crash
with

    ERROR    [Errno 17] File exists: '/home/kstest/.cache/virt-manager/boot'

Fix that by ignoring existing directories instead of explicitly testing
for existence.

The `exist_ok` parameter exists since Python 3.2, and the minimum
supported version is 3.4 now.


crobinso:
Thanks Martin!

PR #179, opened 2020-11-05
virtinst: Fix TOCTOU in pool and volume enumeration

martinpitt:
_fetch_all_pools_raw() and _fetch_vols_raw() have a race condition where
a pool/volume may disappear (from parallel libvirt operations) in
between enumerating and inspecting the objects.

_fetch_vols_raw() already expected that failure in the loop, but not for
the initial storagePoolLookupByName() call.

Also tighten the expected exception: This *should* crash on errors like
AttributeError or SystemExit, just not on dynamic libvirt errors. (Bare
exceptions are highly frowned upon in Python)

https://bugzilla.redhat.com/show_bug.cgi?id=1894359

### DIFF
--- a/virtinst/connection.py
+++ b/virtinst/connection.py
@@ -192,7 +192,16 @@ class VirtinstConnection(object):
     def _fetch_all_pools_raw(self):
         dummy1, dummy2, ret = pollhelpers.fetch_pools(
             self, {}, lambda obj, ignore: obj)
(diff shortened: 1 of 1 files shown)


martinpitt:
I tortured this with the test case in https://bugzilla.redhat.com/show_bug.cgi?id=1894359 and it works fine now.

crobinso:
Thanks Martin!

Issue #181, opened 2020-11-10
Please switch from libappindicator to Ayatana


bigon:
Hello,

virt-manager is using AppIndicator. AppIndicator is deprecated and [Ayatana indicators](https://ayatanaindicators.github.io/) is the replacement for it.

I think that virt-manager should switch to Ayatana indicators or at least it should support both

crobinso:
Thanks for the suggestion. After 10 minutes of googling I can't really tell what the path forward is. Using libappindicator like we currently do works with the gnome-shell extension and KDE so there doesn't appear to be any current functional issue with using libappindicator. But seems like that library is deprecated/unmaintained. If there's an alternative that's usable from python than I'm happy to switch to it or add it as a separate tray implementation.

FWIW In Fedora, `sudo dnf search ayatana` only shows hits for libindicator* which does not appear to do what I want.

Do you know of any python projects that are using ayatana libs? Or providing a statusnotifier icon without using libappindicator?

bigon:
Hello,

Looking on the [debian transition page](https://wiki.debian.org/Ayatana/IndicatorsTransition) it seems that the API is the same.

The only thing that should be change is the fact that you import `appindicator` instead of `appindicator`

Here the patch for another application tries both:
https://loki.unit193.net/cgit/users/unit193/deluge.git/commit/?id=358afd4b33125818f84c3d62bc3a3fe5d53ec67d

bigon:
OK, I've tried the proposed pull request and it seems to work fine

PR #185, opened 2020-11-15
Try to load AyatanaAppIndicator3 in addition of AppIndicator3

bigon:
AyatanaAppIndicator3 is the successor or AppIndicator3 it uses the same API for the time being

Just try to load AyatanaAppIndicator3 in addition of AppIndicator3

Fixes: #181


crobinso:
Thanks! I pushed your commit and added a follow up commit with some minor changes, please try git and verify it still works for your case.

PR #190, opened 2020-11-19
support filterref parameter

gronke:
Add support for network interface filterref.parameter.name and filterref.parameter.value.


phrdina (changes requested):
Thanks for your contribution. There are some changes required to add these two new options, but otherwise looks good.

PR #191, opened 2020-11-20
virtinst: add support for controller target subelement

idryzhov:
Signed-off-by: Igor Ryzhov <iryzhov@nfware.com>


crobinso:
Nice work, thanks!

PR #194, opened 2020-11-24
virtinst: Fix TOCTOU in domain enumeration

martinpitt:
Similar to commit 49a01b5482, _fetch_all_domains_raw() has a race
condition where a domain may disappear (from parallel libvirt
operations) in between enumerating and inspecting the objects.

Ignore these missing domains instead of crashing.

https://bugzilla.redhat.com/show_bug.cgi?id=1901081

-----

I tortured this with the reproducer attached to bugzilla, and I can't get it to crash any more.


crobinso:
Thanks Martin!

PR #195, opened 2020-11-25
spec: fix python macro

idryzhov:
It looks like there is no %{python3} macro in CentOS 8.
Using %{__python3} should be always safe.

Signed-off-by: Igor Ryzhov <iryzhov@nfware.com>


crobinso:
Thank you!

Issue #199, opened 2020-12-01
Testsuite fails with non-English locales


ptoscano:
The test suite fails when run in non-English locales:
```
$ LANG=C pytest
...
= 589 passed in 18.25s =
```
While with Ukrainian (which is generally 100% complete):
```
$ LANG=uk pytest
...
= 133 failed, 454 passed, 2 skipped in 18.11s ==
```
It seems to me the majority of the issues are due to grepping for expected (error) messages of libvirt and virt-manager.

I tried to use the same approach as osinfo-db, i.e. set the locale in the configuration hook of pytest, in `tests/conftest.py`:
```python
def pytest_configure(config):
    # Needed for test reproducibility on any system not using a UTF-8 locale
    locale.setlocale(locale.LC_ALL, 'C')
    locale.setlocale(locale.LC_CTYPE, 'en_US.UTF-8')
```
but unfortunately this does not work because it is executed too late: at that point, other imports of the test suite were done, including libvirt. I tried also to change the locale before other test suite imports, but that did not work either.

Since I ran out of ideas, I'm filing this for reference.

crobinso:
Thanks for the report. Using `os.environ["LANG"] = "en_US.UTF-8"` in `tests/__init__.py` seems to work in my minimal testing. If there's some pytest hook that runs early enough to do that I don't know it offhand

ptoscano:
> Using os.environ["LANG"] = "en_US.UTF-8" in tests/__init__.py seems to work in my minimal testing.

D'oh! So obvious yet I missed it :-/ I'd also unset `LANGUAGE` from the environment, as I see tools setting it as well.

Do you have the change locally already, or should I send a patch/PR?

crobinso:
If you can submit a PR that will help me, we can verify it works in CI too which can be tricky with locales sometimes

PR #200, opened 2020-12-03
tests: force an English UTF-8 locale

ptoscano:
Various checks in the test suite try to match for error/status messages, so using a localized libvirt (or even using translations of an installed version of virt-manager in the system) makes those check fail. Hence, force (using `$LANG`, and unsetting `$LANGUAGE`) an English UTF-8 locale.

Thanks Cole Robinson for spotting the right place where to inject the locale settings.

Fixes: #199

Signed-off-by: Pino Toscano <ptoscano@redhat.com>


crobinso:
Thanks Pino!

PR #233, opened 2021-03-10
Undecorated window

bazsi:
virt-manager: implement undecorated window mode
    
    I have a 4k monitor and I have a VM that uses 1920x1080 resolution, which
    I would like to lay out in the top-right corner of my monitor.
    
    The problem is that as long as the window decoration, menubar and toolbar
    is visible, 1920x1080 does not fit in the 1/4th of my monitor.
    
    Also, the current virtual VGA driver for Windows does not support
    non-standard resolutions like VirtualBox, thus I am either seeing a lot of
    "black" space around my virtual desktop, or I would have to increase
    the size of my virt-manager window.
    
    None of this is appealing, so this patch implements a "windowed" full screen
    mode, which is basically an undecorated window, without a menu or toolbar.
    
    The same overlay is available at the top of the window than what is available
    in full screen.

### DIFF
--- a/ui/vmwindow.ui
+++ b/ui/vmwindow.ui
@@ -183,6 +183,21 @@
                       </object>
                     </child>
                     <child>
+                      <object class="GtkCheckMenuItem" id="details-menu-view-undecorated-window">
+                        <property name="visible">True</property>
+                        <property name="can_focus">False</property>
+                        <property name="label" translatable="yes">Display in _Undecorated Window</property>
+                        <property name="use_underline">True</property>
+                        <signal name="activate" handler="on_details_menu_view_undecorated_window_activate" swapped="no"/>
+                      </object>
+                    </child>
+                    <child>
+                      <object class="GtkSeparatorMenuItem">
+                        <property name="visible">True</property>
+                        <property name="can_focus">False</property>
+                      </object>
+                    </child>
+                    <child>
                       <object class="GtkMenuItem" id="detains-menu-view-size-to-vm">
                         <property name="visible">True</property>
                         <property name="can_focus">False</property>
--- a/virtManager/details/console.py
+++ b/virtManager/details/console.py
@@ -556,12 +556,10 @@ class vmmConsolePages(vmmGObjectUI):
     def _change_fullscreen(self, do_fullscreen):
         if do_fullscreen:
             self._in_fullscreen = True
-            self.topwin.fullscreen()
             self._overlay_toolbar_fullscreen.timed_revealer.force_reveal(True)
         else:
             self._in_fullscreen = False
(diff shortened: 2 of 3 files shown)


bazsi:
This adds a feature described here: https://bugzilla.redhat.com/show_bug.cgi?id=1091311#

crobinso:
Thanks for the patches but I'm sorry I don't think we should carry this in virt-manager. Per [DESIGN.md](https://github.com/virt-manager/virt-manager/blob/master/DESIGN.md) I consider this Advanced Usage, and specifically it violates the section on Advanced VNC/SPICE usage, and there's already a reference there to previously rejecting undecorated windows. I will update that document with the menu bar example too.

There may be other options though:

* windows + qxl video + spice agent does in fact allow arbitrary guest resolutions. I just confirmed with windows 10. Install [virtio-win-guest-tools.exe](https://fedorapeople.org/groups/virt/virtio-win/direct-downloads/archive-virtio/virtio-win-0.1.190-1/virtio-win-guest-tools.exe) and then in virt-manager select `View->Scale Display->Autoresive VM to window`, adjust your window to be the desired size, and the OS will be adjusted to match. You won't get 1920x1080 specifically but the window will fit wherever you want it.

* Simpler but less nice looking: enable `View->Scaling->Always`. Now your VM has 1920x1080 resolution but in your space it will be vertically scaled down a bit. Should still be plenty usable

* For window decorations, there's various external options to hide the title bar from outside the app. xprop and wmctrl like described [here](https://askubuntu.com/questions/906424/remove-decoration-of-single-window-in-gnome-3) worked for me on gnome 3 on x11. I don't know if there's a way to do this with wayland.

* There's not really any way to hide the menu bar that I know of though, sorry.

hockeymikey:
@crobinso What a garbage decision. Who the heck cares about some rigid rules, the user comes first and foremost. Giving the user freedom to decide and use software how best fits them needs to be upheld and championed.

JOJ0:
@hockeymikey Stayin juuust a little more polite might have be the better option, don't you think?

hockeymikey:
@JOJ0 No, I don't think so.

busywhistling:
@bazsi thanks for the PR! I was trying to make this work by building your fork, but couldn't see an Undecorated Window option in the menu. Could you please point me to how I could make this work? (Also since the time you wrote it, v4 of virt-manager is out and I suspect your commits can be integrated with minimal effort. I'd like to look at it, once I make it work for the earlier version)

bazsi:
Well, it's been a long while since I submitted this. Checking out the source there should be a menu item in the top-level menu at right after "Fullscreen" that should toggle the new mode.

To install it I didn't remember having to do too much things, I think I used "setup.py install" to a custom prefix and started from there.

PR #242, opened 2021-04-09
details: enable resizing of displays with new GtkVnc

berrange:
The (forthcoming) 1.2.0 release of GtkVnc introduces support for remote desktop
resize. This is also supported in QEMU >= 6.0.0 when using virtio-gpu.

This introduces support for resize without forcing a new min version of
GtkVnc by just checking for existance of the new API. We don't attempt
to check if the current QEMU instance supports resize, as we gracefully
degrade - the guest simply won't resize and will be rendered as before.

Signed-off-by: Daniel P. Berrangé <berrange@redhat.com>

### DIFF
--- a/virtManager/details/console.py
+++ b/virtManager/details/console.py
@@ -5,8 +5,11 @@
 # This work is licensed under the GNU GPLv2 or later.
 # See the COPYING file in the top-level directory.
 
+import gi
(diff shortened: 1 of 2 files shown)


crobinso:
Cool! The `console_has_agent()` call got mangled so I fixed it. That and added some pylint exclusions with it, and pushed manually

PR #265, opened 2021-06-05
display, graphics: Add default graphics & allow using SPICE on aarch64

fidencio:
This series consists in two simple patches, the first adding default graphics for aarch64, and the second allowing to use SPICE with aarch64.

With this, we have unattended installations working on aarch64 in the exactly the same way it works on x86_64.

This work has has been tested on a NVIDIA Jetson Xavier NX SBC.

### DIFF
--- a/tests/data/cli/compare/virt-install-aarch64-cdrom.xml
+++ b/tests/data/cli/compare/virt-install-aarch64-cdrom.xml
@@ -40,6 +40,18 @@
       <source mode="bind"/>
       <target type="virtio" name="org.qemu.guest_agent.0"/>
     </channel>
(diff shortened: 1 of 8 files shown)


crobinso:
Thanks! But patch 2 is tricky. Particularly RHEL8 qemu only has spice on x86, so this would regress aarch64 installs there. But in the age of qemu modules the existing check could be wrong anywhere nowadays, so I dropped your patch 2 and added a prep patch switching this check to use domcaps reporting. Patch 1 is pushed. Please confirm things still work as expected with real hardware

fidencio:
@crobinso, it does work as expected with real hardware!
Thanks a whole lot!

PR #270, opened 2021-07-02
notify keyboard (un)grabbed

p1gp1g:
There were no way to know if the keyboard is grabbed or not. It has leaded to many errors on my side.
Sending a notification is a way to know, another solution would be with an icon, or a different colour on the systray icon.

This PR is probably not production ready : strings are not in a .ui file and it is not editable in config. And maybe 2 seconds is too much.

Edit: I've tested 1 sec, it is a way better. I'm trying other values

### DIFF
--- a/virtManager/details/viewers.py
+++ b/virtManager/details/viewers.py
@@ -7,6 +7,7 @@
 
 from gi.repository import Gdk
 from gi.repository import GObject
+from gi.repository import Notify
 
 import gi
 gi.require_version('GtkVnc', '2.0')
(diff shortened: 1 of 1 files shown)


crobinso:
Thanks for the patch, but considering we have gotten this far without any keyboard grab notification I don't think it makes sense to add to virt-manager now. This would also be a behavior deviation from the standalone virt-viewer which we try not to do per [DESIGN.md](https://github.com/virt-manager/virt-manager/blob/master/DESIGN.md). *If* we were going to report this though one way to do it would be to alter the window title, like we do when pointer is grabbed for the non-tablet case

PR #275, opened 2021-07-26
cli: --cputune: add emulatorpin.cpuset config

Antagonym:
First attempt at implementing new options. Hope I did it right. Got a few more to add in the near future if this pans out.

### DIFF
--- a/virtinst/domain/cputune.py
+++ b/virtinst/domain/cputune.py
@@ -16,6 +16,14 @@ class _VCPUPin(XMLBuilder):
     vcpu = XMLProperty("./@vcpu", is_int=True)
     cpuset = XMLProperty("./@cpuset")
 
+class _EmulatorPin(XMLBuilder):
+    """
+    Class for generating <cputune> child <emulatorpin> XML
+    """
+    XML_NAME = "emulatorpin"
+
+    cpuset = XMLProperty("./@cpuset")
+
 
 class _CacheCPU(XMLBuilder):
     """
@@ -80,9 +88,11 @@ class DomainCputune(XMLBuilder):
     Class for generating <cpu> XML
     """
     XML_NAME = "cputune"
-    _XML_PROP_ORDER = ["vcpus", "cachetune", "memorytune", "vcpusched"]
+    _XML_PROP_ORDER = ["vcpus", "emulatorpin", "cachetune",
+            "memorytune", "vcpusched"]
 
     vcpus = XMLChildProperty(_VCPUPin)
+    emulatorpin = XMLChildProperty(_EmulatorPin, is_single=True)
     cachetune = XMLChildProperty(_CacheTuneCPU)
     memorytune = XMLChildProperty(_MemoryTuneCPU)
     vcpusched = XMLChildProperty(_VCPUSched)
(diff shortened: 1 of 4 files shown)


virtinst/domain/cputune.py
```diff
@@ -16,6 +16,14 @@ class _VCPUPin(XMLBuilder):
...
     cpuset = XMLProperty("./@cpuset")
 
+class _EmulatorPin(XMLBuilder):
+    """
+    Class for generating <cputune> child <emulatorpin> XML
+    """
+    XML_NAME = "emulatorpin"
+
+    cpuset = XMLProperty("./@cpuset")
+
```
crobinso, line 27: Whitespace issue here: previously the classes were separated by two blanklines, but now it's 1 line before _EmulatorPin and 3 lines after.

virtinst/domain/cputune.py
```diff
@@ -80,9 +88,11 @@ class DomainCputune(XMLBuilder):
     Class for generating <cpu> XML
     """
     XML_NAME = "cputune"
-    _XML_PROP_ORDER = ["vcpus", "cachetune", "memorytune", "vcpusched"]
+    _XML_PROP_ORDER = ["vcpus", "emulatorpin", "cachetune",
+            "memorytune", "vcpusched"]
 
     vcpus = XMLChildProperty(_VCPUPin)
+    emulatorpin = XMLChildProperty(_EmulatorPin, is_single=True)
```
crobinso, line 95: This is completely valid, but for this case it's a bit verbose. `<emulatorpin>` only has one child attribute, and can only be specified once, so it's less code to do:

```
emulatorpin_cpuset = XMLProperty("./emulatorpin/@cpuset")
```

the cli name should stay as `--cputune emulatorpin.cpuset` though. If `<emulatorpin>` ever grows more suboptions, we can split out an _EmulatorPin class like this to organize the code later and it won't affect the command line interface.

virtinst/domain/cputune.py
crobinso: FWIW `./setup.py pylint` pep8 call would have caught this too

crobinso:
Looks pretty good! One whitespace change but I recommended simplifying it a bit too. 

Any future help filling out the command line is definitely appreciated! There's a wiki page here I update periodically with missing cli options: https://github.com/virt-manager/virt-manager/wiki/Unimplemented-domain-XML-command-line-arguments

virtinst/domain/cputune.py
Antagonym: Good to know. I've seen both styles around the rest of the code and I've generally tried following the style what was already there so far, but if that's the preference for "simple" suboptions, I'll stick to that unless there's a lot of suboptions.

virtinst/domain/cputune.py
Antagonym: It did. I just derped and forgot to run it *before* committing ;)

Antagonym:
I've simplified the implementation as suggested, which also neatly eliminates the whitespace issue at the same time.

virtinst/domain/cputune.py
Antagonym: Next up will be iothreadpin[0-9]*. I've already implemented in a branch of my fork as an XMLChildProperty. Considering, it also only has 2 suboptions (`iothread`, and `cpuset), should I simplify that too before opening a pull request?

crobinso:
Thanks! I squashed the two patches together and pushed.

PR #276, opened 2021-07-26
cli: --cputune: add iothreadpin[0-9]* config

Antagonym:
This adds the following suboptions to configure the
\<domain\>\<cputune\>\<iothreadpin\> list:

* iothreadpin[0-9]*.iothread
* iothreadpin[0-9]*.cpuset

\<iothreadpin\> only has two suboptions, but due to its `[0-9]*` nature, I don't think it can be simplified from an XMLChildProperty to an XMLProperty. Feel free to correct me on this if I misunderstood how that mechanism works.


crobinso:
You're correct, XMLChildProperty is required for the case where the subelement can be listed more than once.

PR #278, opened 2021-07-27
cli: --shmem: added support for shared memory devices

Antagonym:
This includes support for the following suboptions:
* `name`
* `role`
* `model.type`
* `size`
* `size.unit`
* `server.path`
* `msi.vectors`
* `msi.ioeventfd`

These should cover all possibilities as specified in the [Domain XML format](https://libvirt.org/formatdomain.html#shared-memory-device). 

Since the current stance on input checking seems to be to keep to a minimum and let libvirt complain about invalid configs, I haven't added any sanity checks beyond some really basic stuff. I have, however, included a list of acceptable values for `<shmem role=X>` and `<shmem><model type=X/>` in the Device class, though they're not enforced. This is analogous to how existing devices I've looked at handled this.

I don't provide any default values, as the specification doesn't either, but if we were to provide any, I'd suggest `size.unit=M` and `size=1` (smallest allowed size), as well as `model.type=ivshmem-plain` (`ivshmem` is deprecated and `ivshmem-doorbell` requires a server, whereas `ivshmem-plain` is stand-alone).

### DIFF
--- a/man/virt-install.rst
+++ b/man/virt-install.rst
@@ -1886,6 +1886,19 @@ Complete details at https://libvirt.org/formatdomain.html#elementsPanic
 
 
 
+``--shmem``
+^^^^^^^^^^^
+
+**Syntax:** ``--shmem`` NAME[,OPTS]
+
+Attach a shared memory device to the guest. The name must not contain ``/`` and must
+not be directory-specific to ``.`` or ``..``
+
+Use --shmem=? to see a list of all available sub options.
+Complete details at https://libvirt.org/formatdomain.html#shared-memory-device
+
+
+
 ``--memdev``
 ^^^^^^^^^^^^
 
--- a/man/virt-xml.rst
+++ b/man/virt-xml.rst
@@ -243,6 +243,7 @@ XML OPTIONS
 * ``--tpm``
 * ``--rng``
 * ``--panic``
+* ``--shmem``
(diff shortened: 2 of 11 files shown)


crobinso:
Nice one! I pushed with a squashed in change to drop the ROLE and TYPE lists: You're right that other device code do similar but they should be similarly changed IMO if the lists or constants aren't explicitly used elsewhere in the code.

Having lists like ROLE_NAMES are really only useful if we then use that list in the virt-manager UI, but even that should be using libvirt domaincapabilities output where appropriate so nowadays I like to avoid that pattern if we can. Sometimes having constants like say DRIVER_NAME_QEMU are useful if we explicitly reference it elsewhere in the code (like in a set_defaults implementation) but that's fairly rare.

And yes you are correct that I prefer to leave validation to libvirt most of the time. Local validation is more code to maintain and often over time our code becomes out of date WRT to libvirt, and then we may reject a config that libvirt would happily except, which is really annoying for users.

crobinso:
For example I see that there's redundant device lists and constants in virtinst/devices/memory.py, I'd take patches to remove those

Antagonym:
Have you already pushed the changes? It doesn't seem to appear in the master branch yet.

crobinso:
My mistake, pushed now

PR #279, opened 2021-07-27
cli: --input: add support for evdev inputs

Antagonym:
This adds support for evdev inputs which were introduced in libvirt 7.4.0,
as well as passthrough inputs and some other misc options to complete
the --input command.

New suboptions:
* source.evdev
* source.dev
* source.repeat
* source.grab
* source.grabToggle
* model

### DIFF
--- a/tests/data/cli/compare/virt-install-many-devices.xml
+++ b/tests/data/cli/compare/virt-install-many-devices.xml
@@ -448,6 +448,13 @@
     <input type="keyboard" bus="usb"/>
     <input type="tablet" bus="usb"/>
     <input type="mouse" bus="ps2"/>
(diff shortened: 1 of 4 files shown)


Antagonym:
The automated test appears to test the code against libvirt 7.0.0, which causes it to fail since support for evdev inputs was only introduced in 0.7.4

crobinso:
I squashed in a bit to only attempt to define the test XML on new enough libvirt, see `predefine_check=` in test_cli.py, and other uses of `pre.*_check=` in that file. Pushed now, thanks a lot!

Antagonym:
Thanks for the tip, this will come in handy for firmware features (introduced in 7.2.0).

PR #280, opened 2021-07-28
cli: --boot: add support of firmware features

Antagonym:
libvirt 7.2.0 introduced support for a list of firmware features
that should or should not be present. Libvirt takes these into
account when auto-selecting a firmware. Currently supported
features are `enrolled-keys` and `secure-boot`.

This also bumps the minimum version for the `singleton-config-2`
test to 7.2.0.

---
Keeping from creating way to many XMLChildProperty with this one caused me some headache until I notice how `<cpu><numa><cell>` is handled using `relative_xpath=""`. Much neater like this!


crobinso:
Looks great, thanks!

PR #281, opened 2021-07-28
cli: --boot: add missing options

Antagonym:
This should fill in all missing suboptions to `--boot` as of 7.5.0.

I can also squash those into a single commit if preferred. I just thought it'd be easier to retrace my steps this way since I've also shuffled bits of code around a lot to make it more readable and easier to understand at a glance which options are related.


virtinst/domain/os.py
```diff
@@ -87,8 +88,13 @@ def _set_bootorder(self, newdevs):
         for d in newdevs:
             dev = self.bootdevs.add_new()
             dev.dev = d
-    bootdevs = XMLChildProperty(_BootDevice)
     bootorder = property(_get_bootorder, _set_bootorder)
+    bootdevs = XMLChildProperty(_BootDevice)
+    smbios_mode = XMLProperty("./smbios/@mode")
+    bootmenu_enable = XMLProperty("./bootmenu/@enable", is_yesno=True)
```
crobinso, line 117: it's a good idea to rename this, enable_bootmenu -> bootmenu_enable, but there's a usage of this in virt-manager too that needs to change as well

virtinst/cli.py
```diff
@@ -2568,6 +2568,7 @@ class ParserBoot(VirtCLIParser):
         "loader.secure": "loader_secure",
         "nvram.template": "nvram_template",
         "smbios.mode": "smbios_mode",
+        "initenv[0-9]*.value": "initenv[0-9]*",
```
crobinso, line 2571: was this alias entry required? typically aliases should only be required when a cli option name is changed, but this is a new option

virtinst/domain/os.py
```diff
@@ -79,12 +79,26 @@ def is_riscv_virt(self):
...
-                       "kernel_args", "dtb", "bootdevs", "smbios_mode"]
+    _XML_PROP_ORDER = [
+            "os_type", "arch", "machine",
+            "loader", "loader_ro", "loader_type", "loader_secure",
+            "firmware", "firmware_features", "nvram", "nvram_template",
+            "bootdevs", "bootmenu_enable", "bootmenu_timeout",
+            "bios_useserial", "bios_rebootTimeout", "smbios_mode",
+            "kernel", "initrd", "kernel_args", "dtb",
+            "acpi_tb", "acpi_tb_type",
+            "init", "initargs", "initenvs", "initdir", "inituser", "initgroup"]
```
crobinso, line 90: I'm okay with reordering things, and comments etc, but if you are going to change options here, it should be to match them with libvirt ordering. Check libvirt [src/conf/domain_conf.c](https://gitlab.com/libvirt/libvirt/-/raw/master/src/conf/domain_conf.c) and search for `virBufferAddLit(buf, "<os");` which is the start of where the `<os>` block is written, follow the order that values are written, which is the order they will show up in the XML, and have our output match that. It's not required for any reason but it makes it easier to diff virt-install XML against libvirt formatted XML which helps in some debugging scenarios

crobinso:
I added some comments but it's looks great overall, thanks this is really helpful!

virtinst/domain/os.py
Antagonym: Fixed & Squashed into the appropriate commit.
Sorry about that. I thought it was fine since the tests all ran through. I'll grep through the code when renaming stuff in the future.

virtinst/cli.py
Antagonym: It's not strictly necessary, no.

I just had a hard time deciding whether to use `.value` or not: `initenv[0-9]*` is shorter and mirrors the underlying XML more closely. On the other hand, `initenv[0-9]*.value` makes more sense semantically and reflects the key=value nature of environment variables better. In the end I chose to value semantics over staying close to the XML (after all, that's one of the reasons to use a CLI tool over editing the XML directly), but kept the other option as an alias as a compromise.

I don't that strongly enough about this, I just thought it was a decent compromise.

virtinst/domain/os.py
Antagonym: Thanks for the tip! Bringing it in line with the XML order was part of the goal, except I didn't know about `domain_conf.c` and ordered mostly as they are presented in the [Domain XML docs](https://libvirt.org/formatdomain.html#operating-system-booting), with a few exceptions.

I've brought it in line with the XML ordering now and squashed those changes into the cleanup/reordering commit.

virtinst/domain/os.py
crobinso: FWIW there are virt-manager UI tests as well: pytest --uitests. But they take over your desktop when being run, require turning on accessibility support, only work on X11, and not many people have run them besides me. They hit most of the code though. Unfortunately dynamic nature of python probably meant this wouldn't have thrown an error anyways

virtinst/cli.py
crobinso: Oh I see now, I should have looked more closely at the XML, I see what you mean.  but I'd prefer in this case to drop the 'value' part of the cli, since that term isn't present in the XML schema. So `--boot initenv0.name=FOO,initenv0=BAR`. Internally it's fine to name it initenv.value though. If everything else looks good I'll make that change before pushing

Antagonym: Alright, it's fixed now.

crobinso:
Thanks!

Issue #204, opened 2020-12-11
Auto-start checkbox in VM gets unchecked after renaming VM


gollum53:
**Distro**: <Ubuntu server 20.04>
Bug: The "Start virtual machine on host boot up" checkbox gets unchecked after renaming VM.

Steps to reproduce:
- Start virtmanager and select a shut-down (not running) virtual machine on another KVM/QEMU hypervisor connected over ssh.
- Open the Show virtual hardware details tab - Boot options and verify that the "Start virtual machine on host boot up" checkbox is checked. If it is not, check it and click Apply.
- Go to Overview, rename the VM, click Apply.
- Close the window and entire virt-manager.
- Run virt-manager again, open the remote hypervisor and see if the checkbox is checked. It is not.

Expected behavior: Leave the checkbox in previous state.

crobinso:
Thanks for the report. Yes this kinda expected given how we implement domain rename, which is basically undefine+define API calls. If we use the virDomainRename API then this will work correctly for the qemu driver but not the other drivers. It shouldn't be too hard to implement fix this for our virt-manager implementation.

PR #283, opened 2021-07-28
using libvirt for renaming domains

coreyberla:
this is the resolution for issue #204 per the discussion in the issue

### DIFF
--- a/virtManager/object/domain.py
+++ b/virtManager/object/domain.py
@@ -540,6 +540,8 @@ class vmmDomain(vmmLibvirtObject):
     ##############################
 
     def rename_domain(self, new_name):
+        if new_name == self.get_name():
+            return
         Guest.validate_name(self.conn.get_backend(), str(new_name))
 
         new_nvram = None
--- a/virtManager/object/libvirtobject.py
+++ b/virtManager/object/libvirtobject.py
@@ -4,7 +4,6 @@
 # This work is licensed under the GNU GPLv2 or later.
 # See the COPYING file in the top-level directory.
 
-import libvirt
 from virtinst import log
 from virtinst import xmlutil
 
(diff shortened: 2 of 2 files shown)


crobinso:
Thanks for the patches, but this is not acceptable as is. That function you are altering is used by storage pools and network objects as well, but you are replacing it with a virDomain specific API. Any domain specific changes should be in domain.py:vmmDomain:rename_domain function.

* Patches should have descriptive commit messages
* Squash fixup patches together with `git rebase -i`
* No unnecessary whitespace changes

I'm curious, is there some behavior of the rename API that you want specifically which virt-manager isn't handling?

coreyberla:
I made the change to resolve the issue in #204 where renaming would lose settings (specifically boot on start). I was able to replicate the issue and resolved it by taking your suggestion on the issue. It was an oversight on my part about the storage and network components. I'll fix and make sure to do so appropriately!

crobinso:
Ah indeed I should have looked at #204 to remind myself.

Switching to use domain rename is one solution but it's also more complicated:
* We can only use it for libvirt drivers that support it.
* We need to keep the old code around for drivers that don't support it.
* We need to make sure we test both code paths in tests/uitests which will take a bit of work

It's a useful thing to do but it's kinda involved. The alternative for #204 is to just implement that specific change in our rename implementation. That's probably as simple as adding

```
oldautostart = self.get_autostart()
# do the rename
self.set_autostart(oldautostart)
```

In libvirtobject.py. That also has the benefit of fixing it for our network and pool rename implementations as well

coreyberla:
Many apologies for the mess of commits (it's been a while since I used git).  I tested your suggestion and it worked, so I added the code.  I also added a check to see if the name actually changed (if the user entered text but ultimately left the name the same, it'd throw an error)

crobinso:
Thanks, I split up your change, tweaked the commit messages, and pushed.

PR #285, opened 2021-08-04
cli: --cpu: implement missing options

Antagonym:
This should add all missing options to `--cpu` as of 7.5.0 and bring the ordering in the XML in line with libvirt's own output.


Antagonym:
Just saw that libvirt v7.6.0 was released a few days ago, but it doesn't seem to add any options here, so this should still be complete as of v7.6.0.

crobinso:
Great work! I rebased and pushed, thanks!

PR #287, opened 2021-08-10
cli: --cputune: implement missing options

Antagonym:
This should add all missing options to --cputune as of 7.6.0 and bring the ordering in the XML in line with libvirt's own output.


crobinso:
Looks great, thanks!

Issue #291, opened 2021-08-25
virt-install error __init__() got an unexpected keyword argument 'text' 


vuzdemav:
This bug appears when running virt-install from the latest git master, where `/usr/bin/python3` is Python 3.6 or lower (e.g. on Ubuntu 18.04 which is where I got this error). It does not appear for the latest release 3.2.0, so it has been introduced since then.

```
Starting install...
ERROR    __init__() got an unexpected keyword argument 'text'
Domain installation does not appear to have been successful.
```

vuzdemav:
After bisecting and investigating the code, I tracked it down to here:

https://github.com/virt-manager/virt-manager/commit/f793986378f84bb409d2451bdb62ca08fd4cb5b4#diff-3fd77cd7003339e4923e8467604ebd1d83f24fcf2031fed499212796b56c67c2R71-R72

```python
        output = subprocess.check_output(cmd,
                stderr=subprocess.DEVNULL, text=True)
```

The `text=` kwarg was added in Python 3.7: https://docs.python.org/3/library/subprocess.html#subprocess.Popen
> Changed in version 3.7: Added the text parameter, as a more understandable alias of universal_newlines. Added the capture_output parameter.

PR #292, opened 2021-08-26
urlfetcher: Make `check_output` call compatible with Python 3.6

vuzdemav:
Fixes: #291

The `text=` keyword was added in Python 3.7 "as a more readable
alias for `universal_newlines=`". This commit switches to the
old form of the name, for backwards compatibility with Python 3.6,
the default Python3 version on Ubuntu 18.04 and RHEL 8.

When support for Python 3.6 is no longer needed, this change can
be reverted.


crobinso:
Thanks for the fix!

Issue #307, opened 2021-10-03
consider cloud.cfg for ssh user (with --cloud-init)


fkooman:
Currently when using --cloud-init the SSH key is configured for the root user. Other cloud-init based systems would use `/etc/cloud/cloud.cfg` (distro specific) to configure a user, e.g. on Fedora this would be the user `fedora`, on CentOS it is user `centos` and on Debian it is user `debian`. 

See e.g. the snippet in `/etc/cloud/cloud.cfg` on Fedora:

```
# System and/or distro specific settings
# (not accessible to handlers/transforms)
system_info:
   # This will affect which distro class gets used
   distro: fedora
   # Default user name + that default users groups (if added/used)
   default_user:
     name: fedora
     lock_passwd: True
     gecos: fedora Cloud User
     groups: [wheel, adm, systemd-journal]
     sudo: ["ALL=(ALL) NOPASSWD:ALL"]
     shell: /bin/bash
   # Other config here will be given to the distro class and/or path classes
   paths:
      cloud_dir: /var/lib/cloud/
      templates_dir: /etc/cloud/templates/
   ssh_svcname: sshd
```

The (big) advantage is that the root user can remain disabled.

fkooman:
Actually there's also a problem with the manpage that is corrected by the linked to PR:

```
       ssh-key=
              Specify a public key to inject into the guest, providing ssh access to the unprivileged account. Example: ssh-key=/home/user/.ssh/id_rsa.pub
```

crobinso:
I'm about to cut a release, so I made the changes I proposed in the PR. We now has `root-ssh-key=PATH` and `clouduser-ssh-key=PATH` options

PR #308, opened 2021-10-03
use default cloud user for SSH key

fkooman:
current `--cloud-init` implementation to `virt-install` uses the
`root` user the provision the SSH key. This patch uses the default
user account as defined in `cloud.cfg` of cloud-init. This matches
behavior of other cloud deployment platforms like e.g. OpenStack.

Possible fix for #307

### DIFF
--- a/virtinst/install/cloudinit.py
+++ b/virtinst/install/cloudinit.py
@@ -76,11 +76,9 @@ def _create_userdata_content(cloudinit_data):
         content += "  expire: False\n"
 
     if cloudinit_data.ssh_key:
-        rootpass = cloudinit_data.get_ssh_key()
+        sshkey = cloudinit_data.get_ssh_key()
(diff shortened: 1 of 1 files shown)


crobinso:
Sorry for the review delay. I agree we should support setting the clouduser ssh key. I think we are already stuck with the old naming though since it's been out for a few releases, and if anyone is using it in scripts we don't want to change the meaning underneath them. What we can do is

* Add `user-ssh-key` or `clouduser-ssh-key` option
* Rename `ssh-key` to `root-ssh-key` but add an unadvertised alias ssh-key for back compat.

What do you think @fkooman ? I can help with the second part if needed

fkooman:
Thanks! I think it is reasonable not to break existing scripts indeed! I'll look at it as soon as I can find some time.

PR #311, opened 2021-10-08
Disk pool volumes should be always attached as 'raw' disks

skobyda:
Usually, when storage volume is attached as a disk and disk xml is filled with
default values, the `<driver type=...>` value is copied from volume's
`<format type=...>`.  This makes sense for volumes of storage pool of type
"dir", where format types include `raw, qcow2...`.

However, the same approach cannot be used for the storage pool of type "disk".
In that case, format types include `none, linux, fat16, fat32...`. Such formats
cannot be used for disk's `<driver type=...>`.

Therefore, when generating disk XML for volume of storage pool type "disk",
driver type should always be "raw".

Documentation: https://libvirt.org/storage.html#StorageBackendDisk


crobinso:
Sorry this stalled, I was offline for a while. I've pushed this now

PR #319, opened 2021-10-26
Handle new format for mdev nodedevs

jonner:
Unfortunately libvirt had to change the nodedev names for mediated
devices due to the fact that mdevctl supports defining multiple mediated
devices with the same UUID as long as there is only one active at a
time. This required changing the nodedev name to ensure uniqueness from
the format 'mdev_$UUID' to 'mdev_$UUID_$PARENT'. This patch allows
virt-install to handle this name format change.

### DIFF
--- a/tests/test_nodedev.py
+++ b/tests/test_nodedev.py
@@ -8,6 +8,7 @@
 import os.path
 
 import pytest
+import libvirt
 
 from virtinst import Guest
 from virtinst import NodeDevice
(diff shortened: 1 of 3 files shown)


jonner:
Oops, pushed that too quickly. Hold off.

jonner:
Strange, I didn't have any test failures when running tests locally.

berrange:
Those checks are from newer libvirt - i've fixed them in this:

https://github.com/virt-manager/virt-manager/pull/320

but waiting for someone to review.

berrange:
If you rebase now it ought to pass tests

tests/test_nodedev.py
```diff
@@ -155,6 +155,15 @@ def testPCIMdev():
...
 
+def testPCIMdevNewFormat():
+    conn = utils.URIs.open_testdriver_cached()
+    devname = "mdev_35ceae7f_eea5_4f28_b7f3_7b12a3e62d3c_0000_06_00_0"
+    dev = _nodeDevFromName(conn, devname)
+    assert dev.name == devname
+    assert dev.parent == "pci_0000_06_00_0"
+    assert dev.device_type == "mdev"
+    assert dev.type_id == "nvidia-11"
+
```
jonner, line 166: So, I just realized that this additional test does not really test what I was trying to test. If I add the following line to the end of the test
```
    assert dev.get_mdev_uuid() == "35ceae7f-eea5-4f28-b7f3-7b12a3e62d3c"
```

then the test will fail if the test is run against libvirt versions earlier than 7.3.0. This is because the `<uuid>` field will not be formatted in the xml description of `dev`, so `dev.get_mdev_uuid()` will return `None`.  I'm not sure about the best way to test this change...

berrange: Can you just set the test to be skipped on older libvirt versions in some way  ?

PR #320, opened 2021-10-29
Fix problems with test suite in new libvirt

berrange:
Libvirt > 7.6.0 introduced uniqueness checks for host devices in

```
commit 9c3b6b7a82373b680a0e4ba4f1baebcb430d45b2
Author: Shalini Chellathurai Saroja <shalini@linux.ibm.com>
Date:   Fri Jun 18 12:46:12 2021 +0200

    conf: verify for duplicate hostdevs
    
    It is possible to define/edit(in shut off state) a domain XML with
    same hostdev device repeated more than once, as shown below. This
    behavior is not expected. So, this patch fixes it.
```

The test suite was mistakenly relying on being able to add the same device twice, but is trivially fixed to use unique devices

### DIFF
--- a/tests/data/cli/compare/virt-install-many-devices.xml
+++ b/tests/data/cli/compare/virt-install-many-devices.xml
@@ -571,15 +571,15 @@
     </hostdev>
     <hostdev mode="subsystem" type="usb" managed="yes">
       <source>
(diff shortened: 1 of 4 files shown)


giuseppe (approved):
LGTM

PR #321, opened 2021-10-29
Explicitly set CPU topology for new VMs and use cores by default

berrange:
This changes the code to always set CPU topology for hvm guests and prefer cores over sockets when doing so. As a side effect of this, we fix many gaps related to CPU dies support.

### DIFF
--- a/man/virt-install.rst
+++ b/man/virt-install.rst
@@ -293,7 +293,8 @@ CPU topology can additionally be specified with sockets, dies, cores, and thread
 If values are omitted, the rest will be autofilled preferring cores over sockets
 over threads. Cores are preferred because this matches the characteristics of
 modern real world silicon and thus a better fit for what guest OS will be
-expecting to deal with.
+expecting to deal with. If all values are omitted, then a default topology
+will be created for fully virtualized machines.
 
 'cpuset' sets which physical cpus the guest can use. ``CPUSET`` is a comma
 separated list of numbers, which can also be specified in ranges or cpus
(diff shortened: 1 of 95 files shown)


crobinso:
I pushed everything except the last 3 patches that turn `<topology>` on by default. I want to play with the UI a bit

crobinso:
Honestly I've been hesitant all these years to make this change, because dealing with `<topology>` in the XML kinda sucks for end users. Any XML editing needs to edit both `<topology>` and `<vcpus>` to match, or libvirt throws an error. virt-manager does the right thing, and we can make virt-xml do the same, but anyone using virsh edit will have to deal with this. Is it a huge issue? No, but it will definitely generate complaints.

I've even more hesitant now since this default is changed in qemu, which means this virt-manager change is going to be largely redundant in 2 years, but all VMs created will be stuck with the `<topology>`, and even new VMs created 5 years from now, unless we eventually revert this. By the time this virt-manager change hits distros, the qemu installed locally will likely already do the right thing.

If libvirt had something like `<vcpus preferCores='yes'>X</vcpus>` that would fill in an implicit `<topology>` at VM startup, this would be simpler IMO.

I'm interested in your thoughts.


Impelementation wise, If we start setting topology by default, I think the virt-manager UI needs to change more. This change will make the current vcpus field largely pointless in the long term, since it will be a readonly spin box and users will be required to interact with the topology UI section.

I'm thinking if the topology is simple, like vcpus = cores or vcpus = sockets, then we keep the topology UI section unexpanded by default, and any interaction with the vcpus field will automagically adjusts the topology as well. So for the default new VM case, the UI will look the same as it does with current git (no default topology), and adjusting total vcpu count works the same way it always has. This is probably worth doing whether we change the default or not

PR #322, opened 2021-11-01
osinfo: use the OS kernel-url-argument if available

ptoscano:
Each OS may specify which kernel argument is needed to specify the
installation source; use it as primary source, falling back to the
current logic. This should help supporting new OSes OOTB.

Signed-off-by: Pino Toscano <ptoscano@redhat.com>

### DIFF
--- a/virtinst/osdict.py
+++ b/virtinst/osdict.py
@@ -593,6 +593,12 @@ class _OsVariant(object):
         Kernel argument name the distro's installer uses to reference
         a network source, possibly bypassing some installer prompts
         """
+        # Let's ask the OS for its kernel argument for the source
+        if hasattr(self._os, "get_kernel_url_argument"):
+            osarg = self._os.get_kernel_url_argument()
+            if osarg is not None:
+                return osarg
+
         # SUSE distros
         if self.distro in ["caasp", "sle", "sled", "sles", "opensuse"]:
             return "install"


berrange:
The change looks reasonable, but I'm wondering if there's need for test suite addition too. I'm not sure if there's currently test suite coverage for this area of code or not - need @crobinso  to give input really.

ptoscano:
There is `test_kernel_url()` in tests/test_osdict.py, and that will now directly fetch the osinfo-db value for the rhel/fedora/sle distros, skipping the own logic in `get_kernel_url_arg()`. I guess that explains the decrease in coverage reported by Codecov.

Ideally, if we know what's the "minimum" osinfo-db version we can assume (even if it is years old), a step after this could be to drop the old logic for distros that provided `<kernel-url-argument>` already in that version.

berrange:
> There is `test_kernel_url()` in tests/test_osdict.py, and that will now directly fetch the osinfo-db value for the rhel/fedora/sle distros, skipping the own logic in `get_kernel_url_arg()`. I guess that explains the decrease in coverage reported by Codecov.
> 
> Ideally, if we know what's the "minimum" osinfo-db version we can assume (even if it is years old), a step after this could be to drop the old logic for distros that provided `<kernel-url-argument>` already in that version.

Hmm, in fact we need to think abuot min  libosinfo API version too.  osinfo_os_get_kernel_url_argument is only added in 1.5.0 - May 2019. We've got 'hasattr' checks for APIs that are way older than 1.5.0, so I think we need one for this too.

ptoscano:
> Hmm, in fact we need to think abuot min libosinfo API version too. osinfo_os_get_kernel_url_argument is only added in 1.5.0 - May 2019. We've got 'hasattr' checks for APIs that are way older than 1.5.0, so I think we need one for this too.

Good point, thanks -- amended.

PR #331, opened 2021-12-02
installerinject: Use root:root instead of +0:+0 as owner

andreabolognani:
Commit c6b5f22fa61d started passing the --owner argument to cpio when injecting files into the initrd to comply with the more strict requirements introduced by systemd starting with Fedora 30.

However, cpio only started accepting the +uid:+gid syntax in version 2.12, which means that the fix actually broke the
initrd inject functionality completely in RHEL 7 and other distros that don't include a recent enough cpio.

Use the user:group syntax instead, which is understood by all versions of cpio, including non-GNU ones.

### DIFF
--- a/virtinst/install/installerinject.py
+++ b/virtinst/install/installerinject.py
@@ -20,7 +20,7 @@ def _run_initrd_commands(initrd, tempdir):
                                  stderr=subprocess.PIPE,
                                  cwd=tempdir)
     cpio_proc = subprocess.Popen(['cpio', '--create', '--null', '--quiet',
-                                  '--format=newc', '--owner=+0:+0'],
+                                  '--format=newc', '--owner=root:root'],
                                  stdin=find_proc.stdout,
                                  stdout=subprocess.PIPE,
                                  stderr=subprocess.PIPE,


berrange:
Do we really still care about RHEL-7 host for virt-manager git master branch ?  Both libvirt and QEMU dropped RHEL-7 already.

Debian 10 has cpio 2.12 so should be ok, as does Ubuntu 18.04 LTS.

andreabolognani:
I've helped a user struggling with this very issue just yesterday.

Overall I fully support the stance that old platforms should stick with old software, but there's a significant difference here: this is not about enabling new features on old platforms but about keeping existing features working. The version of virt-manager shipped with RHEL 7 cannot install anything newer than Fedora 30 because of this issue.

There is also literally zero maintenance overhead added by this change. It simply swaps one syntax for another.

Finally, the +uid:+gid syntax is a GNU-ism that FreeBSD's version of cpio, for example, doesn't implement.

zippy2:
If Andrea's claim about non-GNU cpio is true, then this change makes sense regardless of it fixing RHEL-7.

berrange:
This needs to be reported & fixed against virt-manager in downstream separately, as its not likely to rebase to pull in this fix.

The non-GNU platform motivation is decent justification for this upstream though.

PR #332, opened 2021-12-02
installerinject: Add error detection for initrd injection

andreabolognani:
Any of the commands involved in injecting files into an initrd could fail, and if that happens we should interrupt the installation instead of proceeding as if nothing had happened.

### DIFF
--- a/virtinst/install/installerinject.py
+++ b/virtinst/install/installerinject.py
@@ -43,6 +43,11 @@ def _run_initrd_commands(initrd, tempdir):
     if gziperr:  # pragma: no cover
         log.debug("gzip stderr=%s", gziperr)
 
+    if (cpio_proc.returncode != 0 or
+        find_proc.returncode != 0 or
+        gzip_proc.returncode != 0):  # pragma: no cover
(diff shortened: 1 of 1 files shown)


virtinst/install/installerinject.py
```diff
@@ -95,8 +100,11 @@ def perform_initrd_injections(initrd, injections, scratchdir):
     """
     Insert files into the root directory of the initial ram disk
     """
-    _perform_generic_injections(injections, scratchdir, initrd,
-            _run_initrd_commands)
+    try:
+        _perform_generic_injections(injections, scratchdir, initrd,
+                _run_initrd_commands)
+    except Exception:  # pragma: no cover
+        raise
```
berrange, line 107: What's the benefit of this try/except block - it seems to be functionally identical to what the existing code bit

andreabolognani: You're right, it's not needed because any exception raised by `_perform_generic_injections()` will automatically be visible to the caller of `perform_initrd_injections()`.

PR #341, opened 2021-12-15
Default to UEFI if BIOS is not supported and add a TPM whenever using UEFI

berrange:
Some modern OS will now only support use of UEFI, not legacy BIOS. At the same time they will also generally expect a TPM to be present.

These changes will default to using UEFI if libosinfo declares BIOS is not supported. This relies on a new  API in libosinfo 1.10, so existing installs won't see the new behaviour. It of course also depends on the osinfo-db metadata declaring this fact.

Since use of UEFI generally indicates a desire for a fairly modern platform, it is considered appropriate that the platform should include a TPM by default too. Increasingly OS expect a TPM as a standard, even mandatory, piece of infrastructure. While libvirt has supported swtpm for a while, apps couldn't rely on it being installed by default. So latest libvirt now reports whether tpm is available in domcapabilities, which this code looks for.

The basic point of this is that when selecting a  Windows 11 OS to install, virt-manager/virt-install will now do the right thing to provide a platform that can actually pass the installer's minimum hardware requirements.

### DIFF
--- a/tests/data/capabilities/kvm-x86_64-domcaps-q35.xml
+++ b/tests/data/capabilities/kvm-x86_64-domcaps-q35.xml
@@ -118,6 +118,16 @@
         <value>vfio</value>
       </enum>
     </hostdev>
+    <tpm supported='yes'>
+      <enum name='model'>
+        <value>tpm-tis</value>
+        <value>tpm-crb</value>
+      </enum>
+      <enum name='backendModel'>
+        <value>passthrough</value>
+        <value>emulator</value>
+      </enum>
+    </tpm>
   </devices>
   <features>
     <gic supported='no'/>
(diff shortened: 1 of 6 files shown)


crobinso:
I pushed the first patch, with a tweak to make it abide `--tpm none` instead of `--tpm model=none` which is more inline with what we do with other options.

+1 to the second patch but let's hold off until the win11 entry is in osinfo-db so there's something real to test with

crobinso:
Deleting the master branch seems to have automatically closed this FWIW, it was not intentional on my part

crobinso:
FWIW I pushed your patch, but with some changes on top to be paranoid about this erroring in the future if the API changes.

PR #342, opened 2021-12-15
This removes a bit of legacy cruft around handling osinfo DB

berrange:
This removes the old OS name aliases which were targetted for removal 3 years ago

After doing that a bit of the internal code for osinfo DB handling can be simplified by removing the cache.


crobinso:
Thanks! pushed now

PR #348, opened 2021-12-27
fix small typo

adamaze:


adamaze:
please ignore, it looks like this is a dupe of #339

Issue #353, opened 2022-01-21
Usermode networking is disabled even if QEMU is running as unprivileged


rakbladsvalsen:
Problem:
Userland networking is never enabled, even if QEMU is not running as root. All of this happens under a local `system://` connection (i.e. local libvirtd). 

Some useful info:
```
$ ps aux | grep qemu
kyla      104495  8.9  0.4 9587924 114284 ?      Sl   21:23   0:05 /usr/bin/qemu-system-x86_64 -name guest=win10,debug-threads=on -S -object ...blahblah...

$id
uid=1000(kyla) gid=1000(kyla) groups=1000(kyla),973(libvirt),998(wheel)
```

I am pretty sure this is happening because `[VirtinstConnection.is_privileged()](https://github.com/virt-manager/virt-manager/blob/master/virtinst/connection.py#L361)` is a bit obsolete or it just doesn't take into account connections through a system:// URI.
I was able to fix it with the following snippet within the offending function:
```
        if (
            self.get_uri_path() == "/system" and 
            os.getuid() != 0
        ):
```
This will, however, cause virt-manager to delete all the virtual networks (NAT bridges) at _populate_network_model() because virt-manager somehow assumes unprivileged users cannot use them, even though they perfectly can, as long as they have the permissions to do so (I guess belonging to `libvirt `group is more than enough?)
Removing `[this line](https://github.com/virt-manager/virt-manager/blob/master/virtManager/device/netlist.py#L147)` effectively solves both issues and allows unprivileged users to use virtual networks.

However, I don't see any reason to disable usermode networking (even if QEMU is running as root). It's very convenient if you don't want to mess around installing/setting up dnsmasqd.

**Distro**: Arch Linux. I was also able to reproduce this on master.
**Virt-manager**: 3.2.0-4
**libvirt**: 1:8.0.0-1

PR #354, opened 2022-01-21
WIP: Fix userland networking for unprivileged users

rakbladsvalsen:
This PR cheaply tries to fix usermode networking for unprivileged users. 
For the time being, this is still a WIP as currently a lot of tests will fail since the tests do not expect things like 
`<interface type="user">`. I guess I can add these expected network devices to the XML files, but I'll wait for your input to proceed.

Fixes: https://github.com/virt-manager/virt-manager/issues/353


crobinso:
Thanks, but I put my thoughts in the issue report and I don't think this should change

PR #364, opened 2022-02-24
Add "vmm" shortcut keyword

Ashark:
Adding this keyword will help finding the virtual machine manager application from start menu.


crobinso:
Thanks for the patch. I'm not familiar with how keywords work here, is it just a text search alternative so typing vm or vmm will start matching?

Ashark:
Yes. In KDE you just press Start, then type vmm and you will see the virtual machine manager and can start it from the list. Otherwise you need to type more than three letters to search.

crobinso:
Great, thanks!

PR #385, opened 2022-04-15
add shmem

luc-git:
Hello I have added a ivshmem tab to add it with the gui and not to add it manually on xml

### DIFF
--- a/ui/addhardware.ui
+++ b/ui/addhardware.ui
@@ -314,11 +314,6 @@
                                     <property name="can-focus">False</property>
                                     <property name="has-entry">True</property>
                                     <signal name="changed" handler="on_controller_type_changed" swapped="no"/>
-                                    <child internal-child="entry">
-                                      <object class="GtkEntry">
-                                        <property name="can-focus">True</property>
-                                      </object>
-                                    </child>
                                   </object>
                                   <packing>
                                     <property name="left-attach">1</property>
@@ -478,11 +473,6 @@
                                         <property name="can-focus">False</property>
                                         <property name="halign">start</property>
                                         <property name="has-entry">True</property>
-                                        <child internal-child="entry">
-                                          <object class="GtkEntry" id="combobox-entry3">
-                                            <property name="can-focus">True</property>
-                                          </object>
-                                        </child>
                                       </object>
                                       <packing>
                                         <property name="left-attach">1</property>
@@ -624,11 +614,6 @@
                                     <property name="visible">True</property>
                                     <property name="can-focus">False</property>
                                     <property name="has-entry">True</property>
-                                    <child internal-child="entry">
-                                      <object class="GtkEntry">
-                                        <property name="can-focus">True</property>
-                                      </object>
-                                    </child>
                                   </object>
                                   <packing>
                                     <property name="left-attach">1</property>
@@ -692,9 +677,6 @@
                                         <property name="visible">True</property>
                                         <property name="can-focus">True</property>
                                         <property name="headers-visible">False</property>
-                                        <child internal-child="selection">
-                                          <object class="GtkTreeSelection" id="treeview-selection2"/>
-                                        </child>
                                       </object>
                                     </child>
                                   </object>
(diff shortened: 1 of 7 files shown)


crobinso:
Thanks for the contribution! A brief look of the code looks very thorough!

I can tell glade screwed up addhardware.ui though. Notice in the diff all the internal gtk entrys were removed. This happens when copying chunks of UI around with glade sometimes, I don't understand why exactly. You'll need to adjust your diff to not remove those pieces.

This exposes every shmem config value. Do you use all of those? Generally we aim to have the UI cover common usecases but not every bell and whistle, otherwise it's a lot to test. What config values do you specifically use with ivshmem?

luc-git:
Hello I use ivshmem-plain. I created this entry to simplify those which wants to use looking-glass

crobinso:
It removed child entrys from combo boxes in other parts of the UI. It's a glade weirdness

What is the full XML used for looking glass? Ideally we would just expose enough of the config to cover that use case. Not all the things like MSI bits and things like ivshmem-doorbell which IIRC doesn't have any real users

luc-git:
Hello today I tried to merge my fork to virt-manager (as local repository) and it worked. All entries in addhardware are here and same to comboboxes in children entries. Is it possible to post logs or to give me more infos about your issue? I will try to fix it.

luc-git:
looking-glass needs ivshmem-plain to work

luc-git:
Hello Are you still interested by the pull request? Do I close it or no?

ui/addhardware.ui
```diff
@@ -297,11 +297,6 @@
                                     <property name="visible">True</property>
                                     <property name="can-focus">False</property>
                                     <property name="has-entry">True</property>
-                                    <child internal-child="entry">
-                                      <object class="GtkEntry" id="combobox-entry4">
-                                        <property name="can-focus">True</property>
-                                      </object>
-                                    </child>
```
crobinso, line 304: This deletion here shouldn't happen. This is some kind of artifact of using glade. The instances in this patch don't cause runtime problems as is. But when you open the glade file and then re-save it, at least as of glade-3.38.2-5.fc36.x86_64 which I am using, you'll see diffs like:

```diff
@@ -297,6 +297,11 @@
                                     <property name="visible">True</property>
                                     <property name="can-focus">False</property>
                                     <property name="has-entry">True</property>
+                                    <child internal-child="entry">
+                                      <object class="GtkEntry">
+                                        <property name="can-focus">False</property>
+                                      </object>
+                                    </child>
                                   </object>
                                   <packing>
                                     <property name="left-attach">1</property>
```

Notice that when it refills in the entry, can-focus=False, which will screw up interactivity for those affected entries.
You can fix it with

```console
git show > patch.diff
# edit patch.diff and remove everything but those bad diffs
patch -p1 -R < patch.diff
git commit -a --amend
```

crobinso:
Sorry for the delay. I think adding shmem to the UI is a good idea, but only the options people commonly use. Since looking glass seems to be the primary consumer, let's just target that. So only `name` and `size` fields should be added, everything else should be removed.

If the name needs to be specific for looking glass, we should use a combobox entry and pre-populate the combo box with the special name that looking glass expects. The size value should be a sensible default too, not the smallest possible value.

check `./setup.py pylint` output, there's a small formatting complaint I saw

after that I can help add dogtail UI tests for this, so we have test coverage

luc-git:
Hello after 6 months I fixed my code and I squashed my commits to get a single on.

luc-git:
Is there other things to do?

phrdina:
I'm not convinced we should add this option to UI. The shared memory device is usually located at `/dev/shm/{name}` and by default user doesn't have permissions to use that device as it is `rw-r--r-- qemu:qemu`.

Checking Looking Glass documentation https://looking-glass.io/docs/B6/install/ it already has nice guide how to configure VM in order to use Looking Glass with it.

Adding this into virt-manager will not be enough to make Looking Glass wroking and user will always need to do extra steps on command line with root permissions.

PR #408, opened 2022-06-28
Improve debuggability of virt-install

berrange:
Debugging a recent bug report with virt-install was more challenging than it ought to have been. The only source of information was the virt-install --debug output, because the host OS was an automated system with no direct access available. There were various key bits of info missing in the --debug logs including host capabilities, domain capabilities, virt-install version and various libvirt versions. This series addresses those log output gaps.

### DIFF
--- a/virtinst/capabilities.py
+++ b/virtinst/capabilities.py
@@ -129,6 +129,19 @@ class _CapsGuest(XMLBuilder):
                 ret.append(m.canonical)
         return ret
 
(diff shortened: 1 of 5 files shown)


berrange:
Ping, any comments ?

crobinso:
Thanks, pushed now. I added some changes on top to share the logging between virt-manager and virtinst, otherwise virt-manager was double logging capabilities for example

PR #411, opened 2022-07-09
Pci passthrough

luc-git:
Hello I enhanced pci passthrough by adding: listing iommu group with devices, removing all pci devices in a iommu group, automatically remove all pci devices when iommu is disabled, add rombar importation with gtk filechooser dialog


Quackdoc:
1. seems like add shmem is a left over from another PR you made. does this PR need the other 


2.any reason for the changes from `-` to `_`? and if so are they relevant to the features being added?

I tried to apply this onto master with no luck

PR #412, opened 2022-07-13
enhance pci passthrough support

luc-git:
Hello I enhanced pci passthrough by adding: listing iommu group with devices, removing all pci devices in a iommu group, automatically remove all pci devices when iommu is disabled, add rombar importation with gtk filechooser dialog

### DIFF
--- a/ui/addhardware.ui
+++ b/ui/addhardware.ui
@@ -139,7 +139,7 @@
                                   </packing>
                                 </child>
                                 <child>
-                                  <!-- n-columns=2 n-rows=2 -->
+                                  <!-- n-columns=3 n-rows=3 -->
                                   <object class="GtkGrid" id="table7">
                                     <property name="visible">True</property>
                                     <property name="can-focus">False</property>
@@ -202,6 +202,21 @@
                                         <property name="top-attach">1</property>
                                       </packing>
                                     </child>
+                                    <child>
+                                      <placeholder/>
+                                    </child>
+                                    <child>
+                                      <placeholder/>
+                                    </child>
+                                    <child>
+                                      <placeholder/>
+                                    </child>
+                                    <child>
+                                      <placeholder/>
+                                    </child>
+                                    <child>
+                                      <placeholder/>
+                                    </child>
                                   </object>
                                   <packing>
                                     <property name="expand">True</property>
@@ -244,7 +259,7 @@
                               </packing>
                             </child>
                             <child>
-                              <!-- n-columns=3 n-rows=2 -->
+                              <!-- n-columns=3 n-rows=3 -->
                               <object class="GtkGrid" id="grid5">
                                 <property name="visible">True</property>
                                 <property name="can-focus">False</property>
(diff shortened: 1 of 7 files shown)


Quackdoc:
well it builds, and it does work, however the UI to select the GPU is rather poor. especially on devices with plenty of IOMMU groups, 

I think it would be better to keep the old UI, but use spaces to group the various IOMMU groups. and make adding the entire group optional, 

the new rombar UI is really nice though

luc-git:
I think you have to add all pci devices in a iommugroup otherwise there is "group is not viable" error. So I ensured that all pci devices in iommu group are added and when you remove a pci device all other devices in this iommu group are removed.

luc-git:
Are you one of contributors?

ui/addhardware.ui
```diff
@@ -1,54 +1,57 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- Generated with glade 3.38.2 -->
+<!-- Generated with glade 3.22.2 -->
```
crobinso, line 2: These diffs are huge because you are using an old version of glade, 3.22 was ~4.5 years ago. You need to find something newer.

crobinso:
There's a lot of things happening in this patch at once which is not acceptable for review. You need to split your patches into logically grouped changes so they can be addressed individually.

Adding iommu group awareness to the Add Hardware->PCI is useful. If you want to start with that as a standalone PR, that can be step 1.

Any code that touches the `virtinst/` folder likely needs a unittest. `pytest --cov` shouldn't regress after your changes, so you'll want to check for that.

For rombar path selection, what is your usecase for this? It seems to me like a very advanced feature, and users can use the XML editor if they know they need it. But maybe I'm wrong.

What is the usecase for manually disabling iommu? Is this something you need as part of some workflow, or are you adding it just for completeness? To me it sounds pretty obscure that I don't know if it justifies the maintenance cost to support it in the UI

luc-git:
Rombar selection can be useful for gpus. Sometimes you need to extract the vbios and attach it to the vm to have the signal.

crobinso:
> Rombar selection can be useful for gpus. Sometimes you need to extract the vbios and attach it to the vm to have the signal.

Okay, for something like that which is already very advanced, I don't think it's worth it to maintain in the UI. Users can use the XML editor, or a `virt-xml` command line oneliner

luc-git:
Hello unfortunately I added those lines in the test_nodedev.py file: 

nodename = "pci_104c_ac56"
    obj = _nodeDevFromName(conn, nodename)
    assert obj.is_cardbus_bridge() is True

but the test failed with:

libvirt.libvirtError: Node device not found: no node device with matching name 'pci_104c_ac56'.

Is it necessary to have the Pci device in the PC with which we do the test or not?

The device used in the test is a cardbus bridge device, look at this link:

https://wiki.debian.org/HowToIdentifyADevice/PCI

crobinso:
No, the test suite isn't dependent on host devices being available. It uses mocked libvirt nodedev output stored in `tests/data/testdriver/testdriver.xml`, check that file for other `<device>` examples. Drop your `<device>` XML in there and it should be accessible in `test_nodedev.py` using `open_testdriver_cached()` like other tests use.

phrdina:
Closing this old PR as there are requested changes not addressed, missing commit message and no activity from the author.

PR #415, opened 2022-07-22
ui: Fix too wide lables

flynnjiang:
When some lables have too much content, they will become very wide,
causing the window to widen.

Signed-off-by: Feng Jiang <jiangfeng@kylinos.cn>


crobinso:
Thanks! I pushed this manually after some testing.

PR #422, opened 2022-08-05
Make unicode error when listing data in pool more readable

jkonecny12:
If you unpack tarball from another system it could have a broken names of the files. If this file is part of the pool then running a VM will end up with a hard to understand error. Make the error easy to understand.

Reproducer:
```
mkdir ''$'\320''½'$'\250\316''ļ'$'\376\274\320'
curl -LO https://dl.fedoraproject.org/pub/fedora/linux/releases/36/Cloud/x86_64/images/Fedora-Cloud-Bas
e-36-1.5.x86_64.qcow2
virt-install --connect=qemu:///system --name=test --os-variant=fedora36 --import --disk "size=20,path=$(realpath $PWD)/Fedora-Cloud-Base-36-1.5.x86_64.qcow2"
```

Without this change you will get error:
```
ERROR    Error: --disk size=20,path=/<snip>/virt-manager/test-data2/Fedora-Cloud-Base-36-1.5.x86_64.qcow2: 'utf-8' codec can't decode byte 0xd0 in position 0: invalid continuation byte
```

With this change you will get error:
```
ERROR    Error: --disk size=20,path=/<snip>/virt-manager/test-data2/Fedora-Cloud-Base-36-1.5.x86_64.qcow2: Some files in the 'test-data2' pool have unsupported characters in name. Please remove this file or undefine the pool.
```

This change will make such an error much easy to understand than before. Right now, I would know where to look to solve the issue on my side.

Resolves bug https://bugzilla.redhat.com/show_bug.cgi?id=2115101 .

### DIFF
--- a/virtinst/pollhelpers.py
+++ b/virtinst/pollhelpers.py
@@ -55,7 +55,12 @@ def fetch_volumes(backend, pool, origmap, build_cb):
     typename = "volume"
     list_cb = pool.listAllVolumes
     support_cb = backend.support.conn_storage
-    return _new_poll_helper(origmap, typename, list_cb, build_cb, support_cb)
+    try:
+        return _new_poll_helper(origmap, typename, list_cb, build_cb, support_cb)
+    except UnicodeDecodeError:
+        raise RuntimeError(_("Some files in the '%(pool_name)s' pool have unsupported "
+                           "characters in name. Please remove this file or "
+                           "undefine the pool.") % {"pool_name": pool.name()})
 
 
 def fetch_nodedevs(backend, origmap, build_cb):


virtinst/pollhelpers.py
```diff
@@ -55,7 +55,12 @@ def fetch_volumes(backend, pool, origmap, build_cb):
     typename = "volume"
     list_cb = pool.listAllVolumes
     support_cb = backend.support.conn_storage
-    return _new_poll_helper(origmap, typename, list_cb, build_cb, support_cb)
+    try:
+        return _new_poll_helper(origmap, typename, list_cb, build_cb, support_cb)
+    except UnicodeDecodeError:
+        raise RuntimeError("Some files in the '%s' pool have unsupported "
+                           "characters in name. Please remove this file or "
+                           "undefine the pool." % pool.name())
```
ptoscano, line 63: I think this needs to be translatable, using `_("...")` + `%(foo)s` placeholders.

ptoscano (commented):
Another option here could be to catch the error directly in `_new_poll_helper()`:
https://github.com/virt-manager/virt-manager/blob/8ca7c91697110ae0c1e730fc03237a06d28c5ced/virtinst/pollhelpers.py#L26
possibly providing some sort of textual representation of the problematic item, so the user might have the possibility to fix that item.

virtinst/pollhelpers.py
jkonecny12: Fixed.

jkonecny12:
About moving it into the `_new_poll_helper()`. I don't see a way how to easily show what `pool` is the issue so I don't think it would be that helpful at the end, however, that could be just that I'm unfamiliar with the codebase.

crobinso:
Thanks for the report + patch, especially the reproducer.

The root issue is that libvirt is reporting these wacky filenames as valid storage volumes, when they should really just be ignored. I filed an issue with libvirt: https://gitlab.com/libvirt/libvirt/-/issues/415

Did this come about organically somehow? like is some tool creating these paths? Or was this from some fs experiment you were doing?

I'm guessing this bug has existed in libvirt for a decade, but it's the first time I've heard of anyone hitting it. Unless something has changed that makes this scenario more common, IMO we should just wait for the libvirt fix.

jkonecny12:
I get to this state by downloading something from the internet (windows archive) and having the download folder as libvirt folder. Wasn't intentional a took me while to find out what is wrong.

phrdina:
It would be better to just ignore that volume in `_new_poll_helper()` as Pino suggested but I agree with Cole that unless we get more reports we should just have it fixed in libvirt. So closing this PR for now.

PR #423, opened 2022-08-06
tests: cpio: set owner to 0:0

wegank:
Building virt-manager on macOS with tests enabled fails with the following error:

```
testCLI0051virt_install_initrd_inject: cpio: root:root: invalid group
```

Setting the owner to 0:0 fixes the issue.

### DIFF
--- a/virtinst/install/installerinject.py
+++ b/virtinst/install/installerinject.py
@@ -20,7 +20,7 @@ def _run_initrd_commands(initrd, tempdir):
                                  stderr=subprocess.PIPE,
                                  cwd=tempdir)
     cpio_proc = subprocess.Popen(['cpio', '--create', '--null', '--quiet',
-                                  '--format=newc', '--owner=root:root'],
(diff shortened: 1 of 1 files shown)


bjornfor:
IMHO, that PR message would be good to include in the commit message. (Imagine reading just the commit log.)

ptoscano:
Note the current way, ie `root:root` was done on purpose, see #331. Is there no `root` group on macOS?

/cc @andreabolognani who did #331 some months ago.

andreabolognani:
The `+0:+0` → `root:root` change was done because the former is a GNU-ism, and a somewhat recent at that. Using `0:0` instead sounds sensible to me.

The only reason why I didn't go for that in the first place is that it might lead to unintended results if a user or group with name `0` just so happens to exist on the system, which is why the `+n` syntax was introduced on the GNU side. In practice, it's probably an unlikely enough scenario, where many more tools would break, that it's okay to ignore it.

PR #436, opened 2022-09-16
Packit: initial enablement

lsm5:
This commit enables Packit `copr_build` tasks which will run on every PR and build RPMS using the spec file from Fedora dist-git with Source0 as the archive created from HEAD commit of the PR.

Signed-off-by: Lokesh Mandvekar <lsm5@fedoraproject.org>


lsm5:
@crobinso PTAL

lsm5:
@crobinso gentle re-ping :)

crobinso:
@lsm5 sorry for the delay.

I think we should use the .spec file from virt-manager.git instead? We maintain the spec used for Fedora in virt-manager.git. Otherwise code changes in virt-manager that need matching Fedora changes will fail until Fedora is updated. Is that the intention?

I think we only need to track rawhide too. We quite often don't rebase virt-manager in the stable releases. rawhide is where I expect most breakage to happen anyways. Thoughts?

lsm5:
> @lsm5 sorry for the delay.
> 
> I think we should use the .spec file from virt-manager.git instead? We maintain the spec used for Fedora in virt-manager.git.

That should be doable.

> Otherwise code changes in virt-manager that need matching Fedora changes will fail until Fedora is updated. Is that the intention?

Yes, things could often break if we're using Fedora spec file to build the latest upstream and this could be an early-warning if you want. But to kick things off, we can only care about ensuring that the spec in virt-manager.git is buildable on copr before every PR. 

> 
> I think we only need to track rawhide too. We quite often don't rebase virt-manager in the stable releases. rawhide is where I expect most breakage to happen anyways. Thoughts?

SGTM, I can enable only the rawhide tasks.

lsm5:
@crobinso PTAL

crobinso:
Thanks! Sorry for needing repeated pings

lsm5:
> Thanks! Sorry for needing repeated pings

not at all :) . Thanks a lot for merging this @crobinso . IIRC, you mentioned being ok with including this in all the virt-* repos, is that still the case?

crobinso:
@lsm5 it would be useful for other virt packages, yes, but it may not be an easy fit. A lot of libvirt adjacent packages have CI config that comes from [lcitool](https://gitlab.com/libvirt/libvirt-ci). But if you want to try I suggest raise an issue or PR with [libosinfo](https://gitlab.com/libosinfo/libosinfo) to discuss as a first step, they have a responsive community

lsm5:
> @lsm5 it would be useful for other virt packages, yes, but it may not be an easy fit. A lot of libvirt adjacent packages have CI config that comes from [lcitool](https://gitlab.com/libvirt/libvirt-ci). But if you want to try I suggest raise an issue or PR with [libosinfo](https://gitlab.com/libosinfo/libosinfo) to discuss as a first step, they have a responsive community

Great, thank you @crobinso . I'll check it out.

PR #456, opened 2022-11-10
Check availability of usb redirdevs and spicevmc channels in domcaps

morecache:
v1 -> v2: Follow the original behavior in case of talking to older libvirt.

Libvirt domcaps can advertise the redirect devices and spicevmc channel devices since v8.9.0.
Virt-install adds usb redirdevs and spicevmc channels by default on x86.

This patchset adds checks to avoid using redirect devices/spicevmc channel devices when not supported by the target qemu.


crobinso:
Thanks. I added a patch on top to update test/ domcaps, so we get back to 100% code coverage.

PR #459, opened 2022-11-15
Fix uninitialized controller index

charlesa:
When adding a second virtio-scsi controller to a guest from the Add Hardware dialog an exception is thrown.
Error building device XML: unsupported operand type(s) for +: 'NoneType' and 'int'
This happens because the index for the first controller created was not initialized to zero.

Signed-off-by: Charles Arnold <carnold@suse.com>

### DIFF
--- a/virtManager/addhardware.py
+++ b/virtManager/addhardware.py
@@ -1562,6 +1562,8 @@ class vmmAddHardware(vmmGObjectUI):
         if len(controller_num) > 0:
             index_new = max([x.index for x in controller_num]) + 1
             dev.index = index_new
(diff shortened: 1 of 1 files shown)


crobinso:
Thanks for the patch and the report. It doesn't fix the full issue though. For example, create an aarch64 CDROM install VM, and in customize wizard try to add a single virtio-scsi controller will trigger it too, because virtinst doesn't set index value either. Instead we need to handle non-int index at the usage site.

I fixed it that way in 67832d309 and gave you `Reported-by` attribution

PR #462, opened 2022-11-22
Update Translation

heis2201:
Minor changes.

"Wirtssystem" is the wrong word for host, the correct translation would be "Hostrechner"


ptoscano:
Hi @heis2201,

translations are handled via Weblate, see:
https://github.com/virt-manager/virt-manager/blob/main/CONTRIBUTING.md#translations
So you need to change the translation in Weblate.

Hence, I'm closing this PR, as manually changing .po files will conflict with their handling done by Weblate (which assumes it is the only one modifying them).

Looking forward to see your changes in Weblate!

heis2201:
Alright.

PR #465, opened 2022-12-01
Handle media that matches more than one OS

jonner:
Some ISO files (especially windows versions) don't have a way to
reliably differentiate between different versions. To deal with this,
libosinfo recently added some new APIs to return a list of all OSes that
match a given media source. Switch to these new APIs when available and
only auto-detect the OS if there is no ambiguity (i.e. only one match is
returned).

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2141995

### DIFF
--- a/virtinst/osdict.py
+++ b/virtinst/osdict.py
@@ -18,17 +18,29 @@ from .logger import log
 
 def _process_medialist(medialist):
     nmedias = medialist.get_length()
-    if nmedias == 1:
+    osids = []
+
+    if nmedias == 0:
+        return None
+    elif nmedias == 1:
         return medialist.get_nth(0)
-
-    # can't determine a single definitive match. Print the possible matches and return None
-    if nmedias > 1:
-        ids = []
+    else:
+        # If we can't determine a single definitive match, print the
+        # possible matches and return None
         for n in range(nmedias):
-            ids.append(medialist.get_nth(n).get_os().get_short_id())
-        log.debug("media matches multiple operating systems: %s", ", ".join(ids))
+            # some isos can return multiple matches of the same OS with
+            # different architectures. Ignore duplicates.
+            id = medialist.get_nth(n).get_os().get_short_id()
+            if id not in osids:
+                osids.append(id)
 
-    return None
+        if len(osids) == 1:
+            # just return the first match
+            return medialist.get_nth(0)
+
+        log.debug("media matches multiple operating systems: %s", ", ".join(osids))
+        return None
+
 
 def _media_create_from_location(location):
     if not hasattr(Libosinfo.Media, "create_from_location_with_flags"):
--- a/virtinst/osdict.py
+++ b/virtinst/osdict.py
@@ -16,6 +16,20 @@ from . import xmlutil
 from .logger import log
 
 
+def _process_medialist(medialist):
+    nmedias = medialist.get_length()
+    if nmedias == 1:
+        return medialist.get_nth(0)
+
+    # can't determine a single definitive match. Print the possible matches and return None
+    if nmedias > 1:
+        ids = []
+        for n in range(nmedias):
+            ids.append(medialist.get_nth(n).get_os().get_short_id())
+        log.debug("media matches multiple operating systems: %s", ", ".join(ids))
+
+    return None
+
 def _media_create_from_location(location):
     if not hasattr(Libosinfo.Media, "create_from_location_with_flags"):
         return Libosinfo.Media.create_from_location(  # pragma: no cover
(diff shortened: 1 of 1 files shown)


jonner:
Sorry, I didn't mean to open this PR yet. it's unfinished.

jonner:
Hm, the test is failing because the `tests/data/fakemedia/fake-centos65-label.iso` disk is detected as both i686 and x86_64 architectures.

```
osinfo-detect -a tests/data/fakemedia/fake-centos65-label.iso
Media is bootable.
#1: Media is an installer for OS 'CentOS 6.5 (i686)'
#2: Media is an installer for OS 'CentOS 6.5 (x86_64)'
```

crobinso:
Hmm this is mostly about windows media, correct? Like, it's unlikely an ISO will detect as windows and linux or some totally disparate OSes. So I think in this case I'd rather choose one and log a warning that we matched multiple values.

Thought I think we should filter out when only one is ID is found, but multiple arches reported. I don't think that's anything we care about for virt-install usage, so that shouldn't generate a warning.

Preferring oldest OS value is probably the most conservative option. We could determine that by release date osinfo-db field.

@jonner thoughts?

crobinso:
Can you provide `isoinfo -d -i` output of affected windows media? I can try and mock up some fake iso for the test suite

crobinso:
> Can you provide `isoinfo -d -i` output of affected windows media? I can try and mock up some fake iso for the test suite

@jonnor ping

jonnor:
> > Can you provide `isoinfo -d -i` output of affected windows media? I can try and mock up some fake iso for the test suite
> 
> @jonnor ping

Pong! However, I think it should have been @jonner :D

jonner:
> > Can you provide `isoinfo -d -i` output of affected windows media? I can try and mock up some fake iso for the test suite
> 
> @jonnor ping

Oops, sorry for missing this. Here's what I get for the iso that caused the original downstream issue.

```
$ isoinfo -d -i en-us_windows_server_2022_x64_dvd_620d7eac.iso 
CD-ROM is in ISO 9660 format
System id: 
Volume id: SSS_X64FRE_EN-US_DV9
Volume set id: SSS_X64FRE_EN-US_DV9
Publisher id: MICROSOFT CORPORATION
Data preparer id: MICROSOFT CORPORATION, ONE MICROSOFT WAY, REDMOND WA 98052, (425) 882-8080
Application id: CDIMAGE 2.56 (01/01/2005 TM)
Copyright File id: 
Abstract File id: 
Bibliographic File id: 
Volume set size is: 1
Volume set sequence number is: 1
Logical block size is: 2048
Volume size is: 2710295
El Torito VD version 1 found, boot catalog is in sector 22
NO Joliet present
NO Rock Ridge present
Eltorito validation header:
    Hid 1
    Arch 0 (x86)
    ID 'Microsoft Corporation'
    Key 55 AA
    Eltorito defaultboot header:
        Bootid 88 (bootable)
        Boot media 0 (No Emulation Boot)
        Load segment 0
        Sys type 0
        Nsect 8
        Bootoff 208 520
```

victortoso:
For windows servers, the [osinfo-db!529](https://gitlab.com/libosinfo/osinfo-db/-/merge_requests/529) was needed. Note that in the [tests/isodata/win](https://gitlab.com/libosinfo/osinfo-db/-/tree/main/tests/isodata/win) you can find some isoinfo that is linked with other windows versions (e.g: win2k22) . This was the way that we used to detect isos that could have multiple values in the tests.

virtinst/osdict.py
```diff
@@ -16,6 +16,32 @@
 from .logger import log
 
 
+def _process_medialist(medialist):
+    nmedias = medialist.get_length()
+    osids = []
```
ptoscano, line 21: please reduce the scope of this only where needed (luckily this is not C)

virtinst/osdict.py
```diff
@@ -16,6 +16,32 @@
...
+        return None
+    elif nmedias == 1:
+        return medialist.get_nth(0)
+    else:
+        # If we can't determine a single definitive match, print the
+        # possible matches and return None
+        for n in range(nmedias):
+            # some isos can return multiple matches of the same OS with
+            # different architectures. Ignore duplicates.
+            id = medialist.get_nth(n).get_os().get_short_id()
```
ptoscano, line 33: please use the OS ID rather than the (first) short-id: strictly speaking, libosinfo does not reject the same short-id in different OSes (a test for this was added few months ago in osinfo-db)

virtinst/osdict.py
```diff
@@ -16,6 +16,32 @@
...
+            if id not in osids:
+                osids.append(id)
```
ptoscano, line 35: this sounds like `osids` is a set, so please switch it as such

virtinst/osdict.py
```diff
@@ -16,6 +16,32 @@
...
+    nmedias = medialist.get_length()
+    osids = []
+
+    if nmedias == 0:
+        return None
+    elif nmedias == 1:
+        return medialist.get_nth(0)
+    else:
+        # If we can't determine a single definitive match, print the
+        # possible matches and return None
```
ptoscano, line 29: most likely this comment should be below, right before `log.debug()`?

ptoscano (changes requested):

crobinso:
This has been sitting for a while.  I pushed a commit now that adds `tests/data/fakemedia/fake-win-multi.iso` that will match for multiple windows versions, and I added a test case to make sure virt-install doesn't at least fail when seeing media like that. I'm closing this PR. If anyone wants to send additional patches, please use that fake iso to add test coverage for this stuff.

PR #466, opened 2022-12-12
cloner: clone serial files 

Omrigan:
Before this, on clone the serial file would remain the same for the cloned
domain. This doesn't make much sense as the output would be an intermix of two
unrelated output sequences.

Here we apply the same filename changing algorithm, as with disk files.

### DIFF
--- a/virtinst/cloner.py
+++ b/virtinst/cloner.py
@@ -9,6 +9,7 @@
 
 import re
 import os
+from itertools import chain
 
 import libvirt
 
@@ -20,6 +21,7 @@ from .devices import DeviceInterface
 from .devices import DeviceDisk
 from .logger import log
 from .devices import DeviceChannel
+from .devices import DeviceSerial
 
 
 def _replace_vm(conn, name):
@@ -70,9 +72,9 @@ def _generate_clone_name(conn, basename):
             sep="", start_num=start_num, force_num=force_num)
 
 
-def _generate_clone_disk_path(conn, origname, newname, origpath):
+def _generate_clone_path(origname, newname, origpath, cb_exists):
     """
-    Generate desired cloned disk path name, derived from the
+    Generate desired cloned path for auxiliary files, derived from the
     original path, original VM name, and proposed new VM name
     """
     if origpath is None:
@@ -99,9 +101,7 @@ def _generate_clone_disk_path(conn, origname, newname, origpath):
         clonebase = newname
 
     clonebase = os.path.join(dirname, clonebase)
-    def cb(p):
-        return DeviceDisk.path_definitely_exists(conn, p)
-    return generatename.generate_name(clonebase, cb, suffix=suffix)
+    return generatename.generate_name(clonebase, cb_exists, suffix=suffix)
 
 
 def _lookup_vm(conn, name):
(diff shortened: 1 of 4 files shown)


crobinso:
Thanks for the patch, especially the test coverage!

I agree with the idea. But as implemented, virt-manager's clone wizard won't pick this up. I think `update_serial_files` should be privatized as `_prepare_serial_files` and it should be called from `prepare()`, after `_prepare_nvram()` invocation.

IMO we should unconditionally do this by default, so virtinst/virtclone.py shouldn't need any changes.

`generate_clone_serial_file_name` doesn't need to exist or be static since it's not called anywhere else, so that can be folded into `_process_serial_files`

Thoughts?

Omrigan:
> Thanks for the patch, especially the test coverage!
> 
> I agree with the idea. But as implemented, virt-manager's clone wizard won't pick this up. I think `update_serial_files` should be privatized as `_prepare_serial_files` and it should be called from `prepare()`, after `_prepare_nvram()` invocation.
> 
> IMO we should unconditionally do this by default, so virtinst/virtclone.py shouldn't need any changes.
> 
> `generate_clone_serial_file_name` doesn't need to exist or be static since it's not called anywhere else, so that can be folded into `_process_serial_files`
> 
> Thoughts?

Thanks for taking, I've updated my patches according to your suggestions.

One thing I also noticed is that we need to touch serial file on clone, so that next clone attempts figure out this filename as occupied.

Omrigan:
@crobinso apologies for bother, were you able to take a look?

crobinso:
Hmm, the os.path.exists bits aren't gonna fly, virt-clone can work on remote VMs. We need to be going through the storage APIs for all of this.

The name collision bit may not be much of an issue in practice really, given the way the clone name algorithm works. Did you hit it in real life? Otherwise it may be best to ignore it

Omrigan:
> Hmm, the os.path.exists bits aren't gonna fly, virt-clone can work on remote VMs. We need to be going through the storage APIs for all of this.
> 
> The name collision bit may not be much of an issue in practice really, given the way the clone name algorithm works. Did you hit it in real life? Otherwise it may be best to ignore it

We haven't, let's ignore it, then. I've updated the patches.

virtinst/cloner.py
```diff
@@ -455,6 +458,23 @@ def set_nvram_path(self, val):
...
+
+            def cb_exists(path):
+                # Ignore the check for now
+                return False
+
+            serial.source.path = _generate_clone_path(self.src_name,
+                                                      self.new_guest.name,
+                                                      serial.source.path,
+                                                      cb_exists=cb_exists)
+            Path(serial.source.path).touch()
```
crobinso, line 475: This bit is still making local host changes, even though this could be a remote operation.

Is the `touch` required? In my testing, qemu and/or libvirt will create those log files on VM startup if they don't exist.

crobinso (changes requested):

virtinst/cloner.py
Omrigan: Yes, sorry, that is a leftover for name collision resolution. I've removed it.

crobinso:
Sorry for the delay! Thank you for your patience

Omrigan:
> The name collision bit may not be much of an issue in practice really, given the way the clone name algorithm works. Did you hit it in real life? Otherwise it may be best to ignore it

@crobinso There is actually a case here: when you clone the same VM twice. This I believe could be frequent operation, e. g. you have "template VM" and you create many copies of it. 

What do you think, should I rework the name collision algorithm to consider remote machines and submit another PR?

crobinso:
@Omrigan hmm yeah if you want to give it a shot, I'll take a look. thanks

PR #460, opened 2022-11-18
man: virt-install: document how to disable secure boot

ktdreyer:
Explain how to disable secure boot for a new virtual machine.

PR #467, opened 2022-12-12
virt-install: Improve documentation for UEFI setup

andreabolognani:


andreabolognani:
Idly wondering: would it make sense to implement syntax sugar for handling of Secure Boot?

That is, accept something like `--boot uefi,secure-boot=yes` as a shorthand for `--boot uefi,firmware.feature0.name=secure-boot,firmware.feature0.enabled=yes,firmware.feature1.name=enrolled-keys,firmware.feature1.enabled=yes` and `boot uefi,secure-boot=no` as a shorthand for `--boot uefi,firmware.feature0.name=secure-boot,firmware.feature0.enabled=no`.

Where it gets a bit hairy is that you kinda have two ways to disable Secure Boot: either get libvirt to pick a firmware that contains the feature but doesn't have certificates enrolled, or one where the feature is absent altogether. Depending on the firmware shipped with your distro, you might want one or the other.

So maybe a sort of tristate would be more fitting? `enabled`, `disabled` or `absent`...

We're also considering the possibility of enrolling specific Secure Boot certificates at VM creation time. That could be handled by a second key, along the lines of `secure-boot-enroll=microsoft`, `secure-boot-enroll=redhat`, `secure-boot-enroll=path/to/custom/certs`.

Kinda just throwing ideas at the wall at this point :)

jonner:
Looks good to me.

man/virt-install.rst
```diff
@@ -955,15 +955,28 @@ Some examples:
     Configure the VM to boot from UEFI. In order for virt-install to know the
     correct UEFI parameters, libvirt needs to be advertising known UEFI binaries
     via domcapabilities XML, so this will likely only work if using properly
-    configured distro packages.
+    configured distro packages. This is the recommended UEFI setup.
+
+``--boot uefi,firmware.feature0.name=secure-boot,firmware.feature0.enabled=yes,firmware.feature1.name=enrolled-keys,firmware.feature1.enabled=yes``
+    Configure the VM to boot from UEFI, with Secure Boot support enabled.
```
ktdreyer, line 961:
```suggestion
    Configure the VM to boot from UEFI with Secure Boot support enabled.
```

man/virt-install.rst
```diff
@@ -955,15 +955,28 @@ Some examples:
...
+    Only signed operating systems will be able to boot with this configuration.
+
+``--boot uefi,firmware.feature0.name=secure-boot,firmware.feature0.enabled=no``
+    Configure the VM to boot from UEFI, with Secure Boot support disabled.
```
ktdreyer, line 965:
```suggestion
    Configure the VM to boot from UEFI with Secure Boot support disabled.
```

ktdreyer (approved):
Thanks, I had tried to do this in https://github.com/virt-manager/virt-manager/pull/460 and this is a better change.

andreabolognani:
@crobinso can we please get this merged? It's a simple documentation update and a couple of people already looked at it. Thanks!

crobinso:
Sorry for the delay! Merged now.

andreabolognani:
@crobinso by the way, while I have your attention :) any thoughts about the ideas formulated in [this earlier comment](https://github.com/virt-manager/virt-manager/pull/467#issuecomment-1347098281)? If you think that's something that could be considered, perhaps I can open an issue about it...

crobinso:
> @crobinso by the way, while I have your attention :) any thoughts about the ideas formulated in [this earlier comment](https://github.com/virt-manager/virt-manager/pull/467#issuecomment-1347098281)? If you think that's something that could be considered, perhaps I can open an issue about it...

yes something like that would be nice. it's a bit noisier but `firmware.secure-boot=on|off` is my preference

PR #475, opened 2023-01-05
Remove the pre-defined pcie-root controllers for i440fx in customized dialog

morecache:


crobinso:
I fixed the merge issues locally and pushed now. Sorry for the massive delay. The reason I kept hesitating on just pushing this is because I know it's not a complete fix, there's code architecture issues that make switching q35 back and forth with pc kinda hard to solve. I put my complete thoughts here: https://github.com/virt-manager/virt-manager/issues/720

But this is obviously the biggest issue. Thanks for your work

PR #476, opened 2023-01-06
 console: Select the first embeddable graphics device as graphical console

flynnjiang:
Currently, when there are multiple graphics devices, the first graphics
device is always selected as the graphical console. For the following
configuration:

    <graphics type="egl-headless">
      <gl rendernode="/dev/dri/renderD128"/>
    </graphics>
    <graphics type="spice" autoport="yes">
      <listen type="address"/>
      <image compression="off"/>
      <gl enable="no"/>
    </graphics>

virt-manager will report an error:

    Cannot display graphical console type 'egl-headless'

The patch will iterate through all graphics devices to try to select the
first embeddable graphics device as graphical console.

### DIFF
--- a/virtManager/config.py
+++ b/virtManager/config.py
@@ -208,10 +208,6 @@ class vmmConfig(object):
     def get_ui_dir(self):
         return self.ui_dir
 
-    def embeddable_graphics(self):
-        ret = ["vnc", "spice"]
-        return ret
-
     def inspection_supported(self):
         if not vmmInspection.libguestfs_installed():
             return False  # pragma: no cover
--- a/virtManager/details/console.py
+++ b/virtManager/details/console.py
@@ -691,7 +691,7 @@ class vmmConsolePages(vmmGObjectUI):
             gdevs = self.vm.xmlobj.devices.graphics
             gdev = gdevs and gdevs[0] or None
             if gdev:
-                ginfo = ConnectionInfo(self.vm.conn, gdev)
+                ginfo = ConnectionInfo(self.vm.conn, gdev, 0)
         except Exception as e:  # pragma: no cover
             # We can fail here if VM is destroyed: xen is a bit racy
             # and can't handle domain lookups that soon after
--- a/virtManager/details/console.py
+++ b/virtManager/details/console.py
@@ -221,15 +221,23 @@ class _ConsoleMenu:
         from ..device.gfxdetails import vmmGraphicsDetails
 
         ret = []
+        preferred = None
         for idx, dev in enumerate(devs):
             label = (_("Graphical Console") + " " +
                      vmmGraphicsDetails.graphics_pretty_type_simple(dev.type))
 
-            tooltip = None
             if idx > 0:
                 label += " %s" % (idx + 1)
+
+            tooltip = None
+            if dev.type not in self.embeddable_graphics():
+                tooltip = _("virt-manager cannot display graphical "
+                            "console type '%s'") % (dev.type)
+            elif preferred is not None:
                 tooltip = _("virt-manager does not support more "
                             "than one graphical console")
+            else:
+                preferred = idx
 
             ret.append([label, dev, tooltip])
         return ret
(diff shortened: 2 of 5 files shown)


virtManager/config.py
```diff
@@ -212,6 +212,17 @@ def embeddable_graphics(self):
...
+    def get_perfered_graphics_index(self, gdevs):
+        try:
+            for idx, dev in enumerate(gdevs):
+                if dev.type in self.embeddable_graphics():
+                    return idx
+            if len(gdevs) > 0:
+                return 0
+            return None
+        except:
+            return None
```
crobinso, line 224: This whole function, and the `embeddable_graphics` function, can be moved into console.py actually. `embeddable_graphics` is only called there, it's a remnant from earlier days. Moving 'embeddable_graphics` could be a prep patch

virtManager/config.py
```diff
@@ -212,6 +212,17 @@ def embeddable_graphics(self):
...
 
+    def get_perfered_graphics_index(self, gdevs):
+        try:
+            for idx, dev in enumerate(gdevs):
+                if dev.type in self.embeddable_graphics():
+                    return idx
+            if len(gdevs) > 0:
+                return 0
+            return None
+        except:
```
crobinso, line 223: This except block should be removed. if we screw up the basic code inside, we want to know about it

virtManager/config.py
```diff
@@ -212,6 +212,17 @@ def embeddable_graphics(self):
         ret = ["vnc", "spice"]
         return ret
 
+    def get_perfered_graphics_index(self, gdevs):
```
crobinso, line 215: spelling is `preferred`

virtManager/details/viewers.py
```diff
@@ -134,7 +134,7 @@ def _get_fd_for_open(self):
         if not self._vm.conn.support.domain_open_graphics():
             return None
 
-        return self._vm.open_graphics_fd()
+        return self._vm.open_graphics_fd(self._ginfo.gidx)
```
crobinso, line 137: All this gidx handling could go into the previous patch. Also kill the idx=0 default param in the first patch, force the one user to make it explicit.

virtManager/details/console.py
```diff
@@ -911,7 +917,7 @@ def _console_list_menu_toggled(self, src, dev):
 
     def _populate_console_list_menu(self, ignore=None):
         self._consolemenu.rebuild_menu(
-                self.vm, self._console_list_menu,
+                self.vm, self.config, self._console_list_menu,
```
crobinso, line 920: All this passing `self.config` around shouldn't be required when you move the functions out of config.py. Maybe they would need to be made global or something

crobinso (commented):
Thanks, this looks pretty good! I've got some small comments

virtManager/config.py
flynnjiang: Thanks very much for your suggestions! I have moved `embeddable_graphics` to console.py and removed `get_prefered_grahpics_index`  in v2  after some optimization.

virtManager/config.py
flynnjiang: This function was removed.

virtManager/details/viewers.py
flynnjiang: I've fixed it. Thanks very much!

virtManager/config.py
flynnjiang: This function was removed. Thanks very much!

virtManager/details/console.py
flynnjiang: I have fixed it.

crobinso:
Sorry for the very late response. Code still mostly applies so I gave it a spin.

There's a problem though. There's already some kind of plumbing for listing graphical multiple graphical connections, View->Consoles. However that has the assumption baked in that we only ever try to connect to the first listed graphical device. We need to resolve that somehow, otherwise the UI is out of sync. If I get some time soon i'll try to straighten it out.

crobinso:
> Sorry for the very late response. Code still mostly applies so I gave it a spin.
> 
> There's a problem though. There's already some kind of plumbing for listing graphical multiple graphical connections, View->Consoles. However that has the assumption baked in that we only ever try to connect to the first listed graphical device. We need to resolve that somehow, otherwise the UI is out of sync. If I get some time soon i'll try to straighten it out.

Scratch that, I see you tried to handle it! My testing was bad.

crobinso:
Pushed now, thanks! I'm adding some other cleanups and refactorings on top, if you get some time please confirm your usecase is still working correctly with latest git

flynnjiang:
Your help is greatly appreciated! I just test it with the latest commit (a6b62a20b62ac86e0c40d18f188b312513bafdd9), and it works.

PR #483, opened 2023-01-25
virtinstall: Add hint for using logical_block_size and physical_block_size

fima-s:
virtinstall: Add hint for using logical_block_size and physical_block_size

Add a hint for the user in the form of default values for the logical block and the physical block. Selecting specific values offering a good balance between disk performance and the use of space in primary memory, also these parameters allow you to maintain compatibility with older systems.

Signed-off-by: Efim Shevrin <efim.shevrin@virtuozzo.com>

### DIFF
--- a/virtinst/devices/disk.py
+++ b/virtinst/devices/disk.py
@@ -1015,6 +1015,12 @@ class DeviceDisk(Device):
             if not self.driver_io:
                 self.driver_io = self.IO_MODE_NATIVE
 
+        if self.is_disk():
+            if not self.logical_block_size:
+                self.logical_block_size = 512
+            if not self.physical_block_size:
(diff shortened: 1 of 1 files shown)


crobinso:
Sorry for the very late response.

So this is adding `<blockio logical_block_size="512" physical_block_size="4096"/>` unconditionally to every `<disk device='disk`>`. 

What are the current qemu defaults for those values? Are these values different?
virt-install needs to be very conservative setting defaults and we need to understand the tradeoffs.

Is this relevant to other libvirt drivers? (doesn't look like it from a brief grep of the code). So it would need to be limited to qemu and test driver only.

Do all qemu archs and disk types support this? (seems that way but I didn't dig deeply)

PR #484, opened 2023-01-25
virt-install: improve documentation for secure boot

jonner:
The manpage specifies an old method for disabling secure boot. We should
instead recommend the newer way of automatically choosing a firmware
based on enabled/disabled features.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2112154


jonner:
Oops, Andrea already submitted a better version in #467

PR #472, opened 2022-12-28
Bugfixes

morecache:
https://github.com/virt-manager/virt-manager/issues/469

https://github.com/virt-manager/virt-manager/issues/455

Issue #469, opened 2022-12-22
virt-install ignores `--cloud-init` when using `--location`


larsch:
**Distro**: Arch

**Version**: virt-install 4.1.0

Steps:

```sh
virt-install --osinfo archlinux \
  --location archlinux-x86_64.iso,kernel=arch/boot/x86_64/vmlinuz-linux,initrd=arch/boot/x86_64/initramfs-linux.img \
  --disk size=8 --memory 2048 \
  --cloud-init user-data=user-data,meta-data=meta-data
```

Expected result:

`cloudinit.iso` is generated and included in virtual machine.

Actual result:

Machine is booted without any `cloudinit.iso`.

Notes:

When replacing the `--location` argument with just `--cdrom`, the cloudinit.iso is generated and passed to the VM.

Log: [location-anon.log](https://github.com/virt-manager/virt-manager/files/10286788/location-anon.log)

larsch:
[location-anon.log](https://github.com/virt-manager/virt-manager/files/10286788/location-anon.log)

PR #485, opened 2023-01-26
virtinstall: Fix --location and --cloud-init conflict

lacouture:
So far, --cloud-init is ignored when --location is specified.

Resolves: virt-manager#469

Thanks to @larsch for the initial fix PR #472

This new version of the fix adds a new test case as suggested by @crobinso.
In addition, I've been using this feature repeatedly for the past few hours with no issue with my limited use case: running Ubuntu Server 22.04 autoinstall by passing both cloud-init user-data and the "autoinstall" kernel command line argument.

### DIFF
--- a/tests/data/cli/compare/virt-install-location-iso-and-cloud-init.xml
+++ b/tests/data/cli/compare/virt-install-location-iso-and-cloud-init.xml
@@ -1,0 +1,172 @@
+<domain type="kvm">
+  <name>fedora17</name>
+  <uuid>00000000-1111-2222-3333-444444444444</uuid>
+  <metadata>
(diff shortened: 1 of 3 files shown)


dmlary:
As more support for getting this merged, I just spent four hours debugging why `--cloud-init` was being ignored.  Issue  #469 doesn't have enough links to show up on search engines.  So here's one more until this gets fixed.

`--location` and `--cloud-init` is common in my environment when booting isos with extra kernel args.  Can't use `--cdrom`

ogayot:
I am running into this issue as well when automating Ubuntu installations using virt-install. I can confirm the fix works for my use-case.

In Subiquity (the Ubuntu Server installer), we already have a homemade script wrapping qemu/kvm to perform automated installations. However, having a closer integration with libvirt would be ideal.

Issue #455, opened 2022-11-09
virt-install `--wait -1` should exit gracefuly when VM is destroyed and undefined


skobyda:
Creating a VM with virt-install with infinite waining `--wait -1` and then deleting will result in an exception

**How to reproduce:**
1. Create VM `virt-install --os-variant fedora36 --name myMachine --wait -1 --noautoconsole --cdrom /path/to/iso --debug`
2. Destroy the VM, followed **immediatelly** by undefining it (the issue is a race-condition so it's necessary that there is no timeout between destroy and undefine)
`virsh destroy myMachine; virsh undefine myMachine`

Actual behavior:
virt-install process will exit with exception:
```
Traceback (most recent call last):
  File "/home/skobyda/Work/virt-manager/virtinst/virtinstall.py", line 970, in start_install
    _process_domain(domain, guest, installer,
  File "/home/skobyda/Work/virt-manager/virtinst/virtinstall.py", line 921, in _process_domain
    _wait_for_domain(installer, instdomain, autoconsole, waithandler)
  File "/home/skobyda/Work/virt-manager/virtinst/virtinstall.py", line 892, in _wait_for_domain
    if instdomain.check_inactive():  # pragma: no cover
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/skobyda/Work/virt-manager/virtinst/virtinstall.py", line 838, in check_inactive
    dominfo = self._domain.info()
              ^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/site-packages/libvirt.py", line 1674, in info
    raise libvirtError('virDomainGetInfo() failed')
libvirt.libvirtError: Domain not found: no domain with matching uuid '514b6f69-79de-4021-8b48-148c84afc497' (fedora36-2022-11-9)
```

Expected behavior:
virt-install should exit gracefully, as deleting and undefining VM is nothing unexcpected

skobyda:
Additional notes:

You can achieve the same result if you first undefine the VM:
`virsh undefine myMachine`
and then destroy it:
`virsh destroy myMachine`

crobinso:
PR #486 attempted to fix this with an eventloop, which is certainly the most thorough, but I think that approach needs more work. I think it'd be ok to just check for VIR_ERR_NO_DOMAIN and not throw a backtrace. It's still an error condition though

PR #486, opened 2023-01-27
virt-install: Add lifecycle event loop while installing a persistent …

morecache:
…domain

Currently this patch only handles the UNDEFINED lifecycle event.

We know that undefining a running persistent domain converts it to a transient domain.
During installing a persistent domain, If the domain is undefined & destroyed, The alive virt-install instance exits with an exception because it tries to get info from a non-existent domain.

This patch introduces lifecycle event loop to let virt-install realize the change of domain type (persistent -> transient), So in case domain type changes to transient and is destroyed, The virt-install process can correctly quit rather than try to get the domain info.

Resolves: https://github.com/virt-manager/virt-manager/issues/455

Signed-off-by: Lin Ma <lma@suse.com>

### DIFF
--- a/virtinst/virtinstall.py
+++ b/virtinst/virtinstall.py
@@ -9,6 +9,7 @@ import os
 import sys
 import time
 import select
+import threading
 
 import libvirt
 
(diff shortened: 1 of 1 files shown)


crobinso:
Sorry for lack of response on this.

Adding eventloop support to virt-install is interesting, this bolt on approach is not optimal IMO. virt-install already has a poor mans eventloop with our sleep + dominfo checking. I'd like to see that abstracted out first, probably to a separate file, and then plugging in the proper events support in its place. Maybe even keeping the old code around with an env variable, incase this raises new bugs. I'd want code coverage testing for all of it too

And while this is definitely the most thorough approach to fix the referenced issue, it's also pretty heavyweight. I think there's a simpler fix by tweaking how we handle the VIR_ERR_NO_DOMAIN case.

I'm closing this PR because I think you have moved on, but please rebase+reopen if you are still interested and we can discuss further.

PR #489, opened 2023-01-31
man/virt-install: Add a note about different behavior of --boot on s390x

huth:
It is common on x86 and other architectures to install a guest from network by using "--boot hd,network" with virt-install - as long as the hard disk is not bootable yet, the installation will be started via network, and once it finished, the guest can boot from hd during the next reboot.

However, this does not work on s390x since this architecture only supports one single boot device to be passed to the guest. Thus add a note to the documentation to avoid that people are running again into this common pitfall.

Buglink: https://bugzilla.redhat.com/show_bug.cgi?id=2032472

### DIFF
--- a/man/virt-install.rst
+++ b/man/virt-install.rst
@@ -929,6 +929,8 @@ Some examples:
 ``--boot cdrom,fd,hd,network``
     Set the boot device priority as first cdrom, first floppy, first harddisk,
     network PXE boot.
+    Note: s390x guests only support one boot device, so everything except
+    the first device type will be ignored.
 
 ``--boot kernel=KERNEL,initrd=INITRD,kernel_args="console=/dev/ttyS0"``
     Have guest permanently boot off a local kernel/initrd pair, with the


huth:
Sorry, I initially used the wrong Bugzilla link - fixed it now.

man/virt-install.rst
```diff
@@ -929,6 +929,8 @@ Some examples:
 ``--boot cdrom,fd,hd,network``
     Set the boot device priority as first cdrom, first floppy, first harddisk,
     network PXE boot.
+    Note: s390x guests only support one boot device, so everything except
+    the first device type will be ignored.
```
smitterl, line 933: I tried and both `--boot fd,hd` and `--boot network,hd` boot from hd given with the `--disk` argument.

Maybe change to
```
Note: For s390x guests booting from `cdrom` further boot devices will be ignored.
```

smitterl (changes requested):

man/virt-install.rst
huth: Well, there are no floppy disks on s390x, so "fd" certainly gets ignored. But with `--boot network,hd` the guest only tries to boot via network for me. What's your full command line? ... maybe you've got something else in there that gets into the way?

smitterl: > Well, there are no floppy disks on s390x, so "fd" certainly gets ignored. But with `--boot network,hd` the guest only tries to boot via network for me. What's your full command line? ... maybe you've got something else in there that gets into the way?

I tried several, e.g.

```
virt-install --boot network,hd --disk size=10 --memory 2048 --vcpus 2 --location <install_tree_url> --os-variant rhel9-unknown
```

```
virt-install --boot network,hd --import --disk <bootable_disk_path> --vcpus 2 --memory 2048 --os-variant rhel8-unknown
```

In both cases there is no tftp server available in the default network, I assumed that's why it falls back to disk.

smitterl: Clarified offline: there was a misunderstanding on my side. The doc update refers to setting after installation.

crobinso:
Thanks @huth !

PR #502, opened 2023-03-27
Fixed issue where applications list was too narrow to read in OS info

nose-gnome:
Under OS information the Applications list was too narrow to read, so I have enabled expanding horizontally.

Before:
image

After:
image


crobinso:
Thanks for your work! Sorry for the delay

PR #508, opened 2023-04-17
console: When resizing window to VM, unmaximize only if maximized

DragoonAethis:
Under a KDE Wayland session, unmaximizing a normal window, then resizing it right away results in nothing happening. Check if the console window is maximized first.

As for *why* this is the case, I don't know. If the window is maximized when trying to resize to VM, everything works. Maybe it's some sort of a race condition in Gtk that needs further investigation, but for now, this at least fixes virt-manager.

### DIFF
--- a/virtManager/details/console.py
+++ b/virtManager/details/console.py
@@ -514,7 +514,9 @@ class vmmConsolePages(vmmGObjectUI):
         viewer_alloc = self.widget("console-gfx-scroll").get_allocation()
         desktop_w, desktop_h = self._viewer.console_get_desktop_resolution()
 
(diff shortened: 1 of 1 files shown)


DragoonAethis:
Seems like it's a Gtk quirk as it reproduces on a minimal example too, [filed an issue on Gtk as well](https://gitlab.gnome.org/GNOME/gtk/-/issues/5756).

phrdina:
I did test this on Kubuntu 24.04 with wayland using virt-manager and your test-gtk3.py script, both works correctly. Closing this pull request as I don' think we need this workaround in our code base.

PR #535, opened 2023-06-13
po: Correct zh_CN translation

wszqkzqk:
Correct Chinese translation.


ptoscano:
Hi @wszqkzqk,

translations are handled via Weblate, see:
https://github.com/virt-manager/virt-manager/blob/main/CONTRIBUTING.md#translations
So you need to change the translation in Weblate.

Hence, I'm closing this PR, as manually changing .po files will conflict with their handling done by Weblate (which assumes it is the only one modifying them).

Looking forward to see your changes in Weblate!

Issue #539, opened 2023-07-04
virt-manager does not read requirements of non distro qemu


nicman23:
I have installed a newer qemu that my distro (focal) supports on /opt/. I defined the xml and everything works with libvirtd.

However when i try to connect to it with virt-manager it errors with:

```
Error launching details: Host does not support domain type kvm with machine 'pc-q35-7.1' for virtualization type 'hvm' with architecture 'x86_64'
```

that happens even when the VM is already running. remote-viewer also connects fine on its spice port

lersek:
@crobinso I've started seeing this exact same issue on RHEL-9.2 very recently. Both `virt-manager-4.1.0-4.el9.noarch` and upstream virt-manager at commit 7cd6151a212c throw this exception. It's inexplicable because it used to work absolutely forever.

lersek:
This is definitely a regression in one of: qemu, libvirt, or virt-manager. Previously, I could liberally change the `emulator` element in the domain XML from the RHEL-provided `qemu-kvm` binary to one I compiled myself from upstream qemu, in parallel to changing the machine type accordingly. Previously, both libvirt *and* virt-manager worked well after such an update, but now only libvirt is behaving correctly. With virt-manager, I can still start and shutdown the domain in the summary view, but if I try to open the details view (consoles, details etc) then the above-noted error message is raised. The `virsh capabilities` output indeed does not list the machine types of the custom qemu binary (because the default emulator for the host is the RHEL-provided qemu binary, which has different machine types), but that seems irrelevant: the domain XML already specifies the custom emulator I'd like to use. So the machine type noted in the domain XML should be checked against the particular, specified emulator, not the system default emulator.

lersek:
@berrange @crobinso Right, this is a regression; bisection between `v3.0.0` and 7cd6151a212c ("uitests: Drop hotplug work around, f38 libvirt is fixed now", 2023-05-24) fingers commit 05fcc7410eee ("virtinst: fix caching of domain capabilities", 2022-07-27). Here's the bisection log:

~~~
git bisect start
# status: waiting for both good and bad commits
# bad: [7cd6151a212c4c477a947fe5a7f2b3363dd0dcbd] uitests: Drop hotplug work around, f38 libvirt is fixed now
git bisect bad 7cd6151a212c4c477a947fe5a7f2b3363dd0dcbd
# status: waiting for good commit(s), bad commit known
[...]

PR #558, opened 2023-08-26
virtinst: fix "lookup_capsinfo" usage (Issue 539)

lersek:
These patches address https://github.com/virt-manager/virt-manager/issues/539 ; please see my analysis there (although the commit messages are verbose too).

Thanks!

### DIFF
--- a/virtinst/guest.py
+++ b/virtinst/guest.py
@@ -632,7 +632,12 @@ class Guest(XMLBuilder):
         def _compare_machine(domcaps):
             if self.os.machine == domcaps.machine:
                 return True
-            capsinfo = self.lookup_capsinfo()
+            try:
+                capsinfo = self.lookup_capsinfo()
+            except Exception:
+                log.exception("Error fetching machine list for alias "
+                                  "resolution, assuming mismatch");
+                return False
             if capsinfo.is_machine_alias(self.os.machine, domcaps.machine):
                 return True
             return False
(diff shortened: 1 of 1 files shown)


virtinst/guest.py
```diff
@@ -632,7 +632,12 @@ def lookup_domcaps(self):
         def _compare_machine(domcaps):
             if self.os.machine == domcaps.machine:
                 return True
-            capsinfo = self.lookup_capsinfo()
+            try:
+                capsinfo = self.lookup_capsinfo()
+            except Exception:
+                log.exception("Error fetching machine list for alias "
+                                  "resolution, assuming mismatch");
```
phrdina, line 639: This is not correctly indented.

phrdina:
Otherwise looks good. Probably not ideal using the exception as if we ever get a new place using lookup_capsinfo() we would have to do the exception again but if that happens we can try figure out different solution.

Once the indentation is fixed I'll merge it. Thanks!

virtinst/guest.py
phrdina: Hi Laszlo,

Cannot argue with that as it is indeed consistent. Before this comment I only checked few random multiple line strings where it was indented differently.

So let's keep it like it is now.

lersek:
Thank you, Pavel!

crobinso:
Thanks for handling this guys! I added some unit test coverage in 19b0f3f446ff8fc3b98c676726ee3a42b0b00b74

PR #559, opened 2023-08-27
tests: in xml osinfo url change http to https

Fantu:
this will avoid tests to fails only for different url http->https

this solves a FTBFS that was spotted in debian QA tests and reported in: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1042147


phrdina:
Already fixed as part of https://github.com/virt-manager/virt-manager/pull/558.

ptoscano:
Fixed by commit 6e5c1db6b4a0af96afeb09a09fb2fc2b73308f01, in particular.

PR #564, opened 2023-09-01
virt-clone: Copy disk permissions as well

nertpinx:
When cloning using libvirt APIs the function virStorageVolCreateXMLFrom is used.  However name and permissions are taken from the new XML [1]. By copying the permissions (only the mode is used) we can avoid some unexpected issues.

[1] https://libvirt.org/html/libvirt-libvirt-storage.html#virStorageVolCreateXMLFrom

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2115153


crobinso:
Thanks Martin!

PR #565, opened 2023-09-05
virtinst: On macOS, use root:wheel to run cpio due to root group missing

davidandreoletti:
Problem: macOSs' "root:root" user/group **pair** does not exist. cpio fails to apply the permission as a result. 
Solution: Use macOS's "root:wheel" user/group pair instead.

### DIFF
--- a/virtinst/install/installerinject.py
+++ b/virtinst/install/installerinject.py
@@ -8,6 +8,7 @@ import os
 import shutil
 import subprocess
 import tempfile
+import sys
 
 from ..logger import log
 
@@ -19,8 +20,13 @@ def _run_initrd_commands(initrd, tempdir):
                                  stdout=subprocess.PIPE,
                                  stderr=subprocess.PIPE,
                                  cwd=tempdir)
+
+    owner="root:root"
+    if sys.platform.startswith("darwin"):
+        owner="root:wheel"
+
     cpio_proc = subprocess.Popen(['cpio', '--create', '--null', '--quiet',
-                                  '--format=newc', '--owner=0:0'],
+                                  '--format=newc', f'--owner={owner}'],
                                  stdin=find_proc.stdout,
                                  stdout=subprocess.PIPE,
                                  stderr=subprocess.PIPE,


ptoscano:
This was changed already a couple of times in the past: PR #423, PR #331; the last change was explicitly mentions macOS. So:
- I don't think we should switch again from `0:0`, at least on non-macOS
- if the previous/last fix was done for macOS and it broke again, please specify how the previous change was not effective, and what's the current situation

davidandreoletti:
@ptoscano 

On macOS, the **root:root** pair (as in **user_name:group_name** pair) does NOT exist, yet the pair **0:0** (as **user_id:group_id**) does exist and maps to **root:wheel**. 

#423 sets owner to _0:0_, regardless of platform. Prior to cpio 2.12, user:group group syntax is understood as _user_id:group_id_ OR _user_name:group_name_. On platform, where 0:0 represents _user_name:group_name_ this is an issue.
#331 sets owner to _+0:+0_, regardless of platform, to overcome the ambiguity. Prior to cpio 2.12, +user_id:+group_id is not supported. Post cpio 2.12, _+0:+0_ is only understood as "user_id:group_id"

The proposed fix changes the owner on the macOS platform only and is unambiguous regardless of the cpio version (ie no new syntax and root:wheel cannot be consider as a  user_id:group_id. 

Note: The _+user_id:+group_id_ syntax was introduced in 2014-01-30 and release in cpio 2.12 in 2015.

crobinso:
@davidandreoletti Thanks for the analysis but I'm still a bit confused. Is this functionally fixing anything for you? Doesn't current `--owner=0:0` do the same thing as `--owner=root:wheel` on MacOS, unless there's a user and/or group named `0` ?

davidandreoletti:
> @davidandreoletti Is this functionally fixing anything for you? 
Yes, I have to use the proposed fix to get vert-install to successfully "cloud-init" a fedora machine, on my current version of macOS.

> Doesn't current `--owner=0:0` do the same thing as `--owner=root:wheel` on MacOS, unless there's a user and/or group named `0` ?

0:0 would probably work on macOS (not tried) but as I said best to use an **unambiguous** pair. Hence root:wheel for macOS.

davidandreoletti:
> Doesn't current --owner=0:0 do the same thing as --owner=root:wheel on MacOS, unless there's a user and/or group named 0 ?

On my machine virt-manager 4.1.0_4 is installed ([homebrew package](https://github.com/Homebrew/homebrew-core/blame/3791be30acb98afee625099af78dfb5986f91258/Formula/v/virt-manager.rb)) uses `owner=root:root` but on the main branch now it is "0:0". 

Let me try with 0:0 first and get back to you.

davidandreoletti:
@ptoscano 

https://github.com/virt-manager/virt-manager/releases/tag/v4.1.0 was released on August 2022! 

The `0:0` bug fix was made on (Dec 2022)](https://github.com/virt-manager/virt-manager/commit/75a25e37660c5578587f4a7a75917cf98d77cf7e) and therefore not available on homebrew based installations.

Would you mind producing a new virt-manager please (I will submit a patch to home-brew to pickup the new release) ? 

This PR is now moot. It will be closed as soon as a new release is out.

davidandreoletti:
> Thanks for checking. Yes we are overdue for a release. I'm trying to make time to catch up with the backlog

What's the likely timeframe for a new release ?

davidandreoletti:
> What's the likely timeframe for a new release ?

@crobinso (ping)

ptoscano:
@davidandreoletti I suggest you to submit commit 75a25e37660c5578587f4a7a75917cf98d77cf7e as separate patch tohomebrew for now, so at least this will fix `virt-install` on macOS.

> This PR is now moot. It will be closed as soon as a new release is out.

I'd rather close this now than letting it linger in the PR list among other potential PRs to review/merge. Would it be OK for you?

davidandreoletti:
Happy you proposed something but the proposed solution is not ok because:
- they are up to 50 others upstream contributions still missing.
- no release timeframe to eventually stop custom homebrew patches "shenigans".

ptoscano:
>     * they are up to 50 others upstream contributions still missing.

I'm not sure how the number of commits upstream since the last release is relevant to this issue. The fix needed for making the installation of Linux guests on macOS working is only one of them, and it does not depend on any other change that was done after the last stable release.

>     * no release timeframe to eventually stop custom homebrew patches "shenigans".

Distros do backport fixes needed to make things working all the time, no matter whether there is a planned release or not. In this case, it would not be "crazy downstream patching", but rather backporting a simple one-liner fix from upstream.

I appreciate the concern about having a new release, and I fully understand it (I'm the Debian packager of virt-manager, so I know this very well). OTOH, if there is anything that breaks virt-manager bad enough (like the invalid `--format` parameter for `cpio` on macOS), then I'd rather backport that even in case there is no release on sight.

ptoscano:
> > Distros do backport fixes needed to make things working all the time, no matter whether there is a planned release or not.
> 
> Yes they do. Is not it mostly because they can't bump the software version too much between named releases ? (Otherwise non testing Debian would have much more recent software versions :-))

Not correct: they do backport patches also to unstable & testing. Sometimes you need to backport patches to fix a build issue for a source that blocks other packages, even if a new release of a source is planned for the next month (and yes, I've done that various times).

> > In this case, it would not be "crazy downstream patching", but rather backporting a simple one-liner fix from upstream.
> 
> My concern is not this fix in particular but how many more may be needed (unknown).

I find this a bit overly speculating, and I don't think it helps.

> You are free to do as you please with this issue now.

I'll close this, since it is about fixing a bug which was already fixed.

Thanks anyway for the patch.

crobinso:
> > What's the likely timeframe for a new release ?
> 
> @crobinso (ping)

The unfortunate answer is, when I get to it. Usually when I make time predictions they are way off.

PR #583, opened 2023-09-17
Add support for hvf domain type

mohd-akram:
The default "qemu" domain type on macOS is very slow, so pick up "hvf" when it's available.

### DIFF
--- a/virtManager/connection.py
+++ b/virtManager/connection.py
@@ -220,6 +220,8 @@ class vmmConnection(vmmGObject):
             label = "QEMU TCG"
         elif domtype == "kvm":
             label = "KVM"
+        elif domtype == "hvf":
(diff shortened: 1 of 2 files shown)


crobinso:
Thanks for the patch. We will want unit test coverage for this too though. I can handle that piece but I need some info from you:

Can you attach output of these from your host:

* `virsh capabilities`
* `virsh domcapabilities --virttype hvf`

mohd-akram:
Sure thing, I've attached them.

[capabilities.txt](https://github.com/virt-manager/virt-manager/files/12664972/capabilities.txt)
[domcapabilities.txt](https://github.com/virt-manager/virt-manager/files/12664974/domcapabilities.txt)

PR #587, opened 2023-09-26
Add 'sw_64' cpu support

KunWuChan:
add sw_64 cpu support for 'Shenwei/Sunway CPU architecture'

### DIFF
--- a/virtManager/createvm.py
+++ b/virtManager/createvm.py
@@ -859,8 +859,6 @@ class vmmCreateVM(vmmGObjectUI):
 
         defmachine = None
         prios = []
-        if self._capsinfo.arch in ["sw_64"]:
-            defmachine = "core3"
 
         recommended_machine = virtinst.Guest.get_recommended_machine(
                 self._capsinfo)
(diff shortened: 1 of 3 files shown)


virtManager/createvm.py
```diff
@@ -858,6 +859,9 @@ def _populate_machine(self):
 
         defmachine = None
         prios = []
+        if self._capsinfo.arch in ["sw_64"]:
+            defmachine = "core3"
```
ptoscano, line 863: i think this should be done in `virtinst.Guest.get_recommended_machine()`, in the internal `_qemu_machine()` function of it: see the existing cases

KunWuChan: Thanks for the kind reminder, I've fixed it

virtManager/createvm.py
```diff
@@ -858,6 +859,7 @@ def _populate_machine(self):
 
         defmachine = None
         prios = []
+
```
ptoscano, line 862: extra/unneeded change

ptoscano (commented):
other than the things I already pointed out, I'm not familiar enough with these CPU-related bits to know whether these "sw_64" changes are OK; I'll leave this to @crobinso

crobinso:
This seems to be about non-upstream qemu support. Before putting code into virt-manager we need upstream qemu and libvirt to support your feature. After sw_64 support is upstream, please re-file this PR. Thanks!

PR #591, opened 2023-10-10
console: allow localhost connections over TCP

enool:
Fix console config validation for an edge case where both libvirt and Spice/VNC TCP connection are on localhost. Transport over TCP does not necessarily mean that libvirt connection is to a remote host.

Compare libvirt and graphics device addresses to localhost individually. Raise an error only when guest device is bound to localhost but libvirt connection is non-local (remote).

Validation that prevents fully local TCP seems to go back all the way to bc13c302de4 ("console: Warn if qemu+tcp URI and listen == 127.0.0.1").

### DIFF
--- a/virtManager/details/sshtunnels.py
+++ b/virtManager/details/sshtunnels.py
@@ -69,7 +69,8 @@ class ConnectionInfo(object):
 
         if (not self.need_tunnel() and
             self.transport and
-            self._is_listen_localhost(self.get_conn_host()[0])):
+            self._is_listen_localhost() and
+            not self._is_listen_localhost(self._connhost)):
             return _("Guest is on a remote host with transport '%s' "
                 "but is only configured to listen locally. "
                 "To connect remotely you will need to change the guest's "


firstdorsal:
I have the same problem. Would love to see that merged!

phrdina:
I'm not convinced we need this patch merged. If I understand correctly you are using virt-manager with custom URI connection looking like `qemu+tcp://localhost/system` and your VM is configured to listen on `localhost`. I don't understand why would someone use configuration like that and why not connecting dirrectly to `qemu:///system`.

Can you please describe your use-case?

enool:
Hello!

The main use case is indeed to connect to several different local Libvirt containers. Virt-viewer works fine for this but switching between the two is a bit annoying. Is there anything I can do to make this easier to merge?

See also https://github.com/virt-manager/virt-manager/issues/592

phrdina:
That sounds reasonable and would be nice to have it fixed. However, I don't think the patch is correct, using `_is_listen_localhost` to check the `_connhost` is not what the function should do. In case the `_connhost` is not IP address the code inside the function will create exception that we will catch and return `False`. So it works but doesn't feel right using it like that. We need to figure out different way how to check it.

crobinso:
Internally we already convert connhost == 'localhost' to '127.0.0.1', so this will work for libvirt URIs pointing to `localhost` and `127.0.0.1` but not `myfqdn`, but I think that's ok.

I ran this through the uitest suite which already has a lot of coverage here, and didn't hit any regressions, so I'm merging!

I'll also push an extra uitest to make sure we don't regress. 

Thanks for your patience!

PR #598, opened 2023-10-23
Fix filesystem socket.source

jonner:
When specifying the socket.source option for filesystem devices, like
this:
  --filesystem type=mount,driver.type=virtiofs,source.socket=/xyz.sock,target.dir=tag1

virt-install is writing the xml as:

    <filesystem type="mount">
      <source>
        <socket>/xyz.sock</socket>
      </source>
      <target dir="tag1"/>
      <driver type="virtiofs"/>
    </filesystem>

This produces an error such as:

  ERROR missing source information for device mount_tag1

But the socket should be an attribute of source rather than a child
element. After this patch, the same command results in the following XML
and no error is produced:

    <filesystem type="mount">
      <source socket="/xyz.sock"/>
      <target dir="tag1"/>
      <driver type="virtiofs"/>
    </filesystem>

Resolves: RHEL-1126

### DIFF
--- a/tests/data/cli/compare/virt-install-many-devices.xml
+++ b/tests/data/cli/compare/virt-install-many-devices.xml
@@ -509,9 +509,7 @@
       <readonly/>
       <space_hard_limit>1234</space_hard_limit>
       <space_soft_limit>500</space_soft_limit>
-      <source pool="pool1" volume="vol">
-        <socket>/tmp/foo.sock</socket>
-      </source>
+      <source pool="pool1" volume="vol" socket="/tmp/foo.sock"/>
       <target dir="/foo"/>
       <binary path="/foo/virtiofsd" xattr="off">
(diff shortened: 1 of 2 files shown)


ptoscano (approved):
This matches the libvirt documentation:
https://libvirt.org/formatdomain.html#filesystems

> `source`
> 
> The resource on the host that is being accessed in the guest. The `name` attribute must be used with `type='template'`, and the `dir` attribute must be used with `type='mount'`. For `virtiofs`, the `socket` attribute can be used to connect to a virtiofsd daemon launched outside of libvirt. In that case, the `target` element does not apply and neither do most virtiofs-related options, since they are controlled by virtiofsd, not libvirtd. The `usage` attribute is used with `type='ram'` to set the memory limit in KiB, unless units are specified by the `units` attribute.

LGTM, thanks.

PR #609, opened 2023-12-19
Update diskbackend.py

anonymousdouble:
refactor with chain constant value assignment to make code more Pythonic, concise and efficient.


crobinso:
Thanks for the suggestion but this is not an acceptable change IMO. Main reason is it doesn't match the style used elsewhere in the codebase. If we were going to start using this style I would want it to be consistently applied tree wide and probably enforced with a tool like `black` or similar, but that requires it own discussion first.

Issue #628, opened 2024-02-22
Add loongarch cpu support for virt-manager


lixianglai:
LoongArch is the general processor architecture of Loongson.
You can follow the links below to learn more information related to loongarch:
https://en.wikipedia.org/wiki/Loongson#LoongArch
https://github.com/loongson/LoongArch-Documentation

At present, loongarch is committed to the development of virtualization and open source work.

The latest version of qemu has been very good support loongarch architecture, You can use qemu tcg to emulate loongarch.

The libvirt related patch supporting loongarch has been submitted to the third version and will soon be accepted by the libvirt community. 

Therefore, we plan to submit virt-manager related patches supporting loongarch to the community so that the open source work of loongarch virtualization is complete.

  You can get the virt-manager code with loongarch patch from the link below:
https://github.com/lixianglai/virt-manager
branch: loongarch

The libvirt-related patch is not yet accepted by the community, but you can get the corresponding patch from the following link：
https://gitlab.com/lixianglai/libvirt
branch: loongarch

  loongarch's virtual machine bios is not yet available in qemu,So you need to compile loongarch UEFI yourself,
you can refer to the following link to compile UEFI:
https://github.com/tianocore/edk2-platforms/blob/master/Platform/Loongson/LoongArchQemuPkg/Readme.md

Here we provide compiled UEFI for ease of testing,you can get it from the following link:
https://github.com/lixianglai/LoongarchVirtFirmware
(Note: You should clone the repository using git instead of downloading the file via wget or you'll get xml)
We named the bios QEMU_EFI.fd, QEMU_VARS.fd is used to store pflash images of non-volatile variables.
After installing qemu-system-loongarch64,You can install the loongarch bios by executing the script
install-loongarch-virt-firmware.sh 

  Since there is no fedora operating system that supports the loongarch
architecture, you can find an iso that supports loongarch at the link below for testing purposes:
  https://github.com/fedora-remix-loongarch/releases-info
The operating system provided above may have minor problems of one kind or another,
but you can also find out from the following link that openEuler's operating system supports loongarch:
https://muug.ca/mirror/openeuler/openEuler-22.03-LTS/ISO/loongarch64/

  Well, if you have completed the above steps I think you can now install loongarch virtual machine,
you can install it through the virt-manager graphical interface, or install it through vrit-install,
here is an example of installing it using virt-install:
virt-install  \
--virt-type=qemu \
--name  loongarch-test \
--memory 4096 \
--vcpus=4 \
[...]

PR #629, opened 2024-02-26
Add loongarch cpu support for virt-manager

lixianglai:
The virt-manager code is currently not perfect and lacks corresponding test cases. However, according to the suggestion, I first create a Pr for everyone to review. The following is a more detailed description：

https://github.com/virt-manager/virt-manager/issues/628


virtinst/devices/disk.py
```diff
@@ -979,6 +979,8 @@ def _default_bus(self, guest):
             return "sd"
         if guest.os.is_q35():
             return "sata"
+        if self.is_cdrom() and guest.os.is_loongarch64():
+            return "scsi"
```
crobinso, line 983: this line isn't hit by the test case. have one of your test_cli.py cases add a `--disk device=cdrom` and `pytest --regenerate-output` to update the test content

virtinst/guest.py
```diff
@@ -1022,7 +1028,7 @@ def _add_default_graphics(self):
         if self.os.is_container() and not self.conn.is_vz():
             return
         if (not self.os.is_x86() and
-            not self.os.is_pseries()):
+            not self.os.is_pseries() and not self.os.is_loongarch64()):
```
crobinso, line 1031: please preserve the same formatting here and put this on a new line

virtinst/guest.py
```diff
@@ -1164,6 +1170,8 @@ def _add_spice_sound(self):
         self.add_device(dev)
 
     def _add_spice_usbredir(self):
+        if (self.os.is_loongarch64()):
+            return
```
crobinso, line 1174: also, why disable this? does libvirt or qemu throw an error? if not IMO we should keep it enabled

crobinso (changes requested):

virtinst/devices/disk.py
lixianglai: Sorry, I did not do the above operation in test_cli, but I think it is unnecessary to add these statements, Because it has the following statement on it：

`
if (self.is_cdrom() and
     guest.supports_virtioscsi() and
     not guest.os.is_x86()):
     return "scsi"
`
So I will delete the previously added statement when I push patch next time.

tests/data/capabilities/kvm-loongarch64-domcaps.xml
```diff
@@ -0,0 +1,167 @@
...
+  <domain>kvm</domain>
+  <machine>virt</machine>
+  <arch>loongarch64</arch>
+  <vcpu max='256'/>
+  <iothreads supported='yes'/>
+  <os supported='yes'>
+    <enum name='firmware'/>
+    <loader supported='yes'>
+      <value>/obviously/fake/firmware1.fd</value>
+      <value>/obviously/fake/firmware2.fd</value>
```
andreabolognani, line 12: This looks like it comes right out of the libvirt test suite. Here's what the `<os>` element looks like on my running installation:

```xml
<os supported='yes'>
  <enum name='firmware'>
    <value>efi</value>
  </enum>
  <loader supported='yes'>
    <value>/usr/share/edk2/loongarch64/QEMU_EFI.fd</value>
    <enum name='type'>
      <value>rom</value>
      <value>pflash</value>
    </enum>
    <enum name='readonly'>
      <value>yes</value>
      <value>no</value>
    </enum>
    <enum name='secure'>
      <value>no</value>
    </enum>
  </loader>
</os>
```

Notice EFI support being advertised and fake paths being replaced with real ones.

tests/data/cli/compare/virt-install-kvm-loongarch64-basic.xml
```diff
@@ -0,0 +1,93 @@
...
+  <uuid>00000000-1111-2222-3333-444444444444</uuid>
+  <metadata>
+    <libosinfo:libosinfo xmlns:libosinfo="http://libosinfo.org/xmlns/libvirt/domain/1.0">
+      <libosinfo:os id="http://fedoraproject.org/fedora/19"/>
+    </libosinfo:libosinfo>
+  </metadata>
+  <memory>65536</memory>
+  <currentMemory>65536</currentMemory>
+  <vcpu>2</vcpu>
+  <os>
```
andreabolognani, line 12: The fact that `<os firmware='efi'>` is not used makes this looks quite unrealistic. Might be connected to the lack of EFI information in the domcapabilities?

tests/data/cli/compare/virt-install-kvm-loongarch64-basic.xml
```diff
@@ -0,0 +1,93 @@
...
+    </channel>
+    <channel type="spicevmc">
+      <target type="virtio" name="com.redhat.spice.0"/>
+    </channel>
+    <input type="tablet" bus="usb"/>
+    <input type="keyboard" bus="usb"/>
+    <graphics type="spice" port="-1" tlsPort="-1" autoport="yes">
+      <image compression="off"/>
+    </graphics>
+    <sound model="ich6"/>
```
andreabolognani, line 41: `ich6` is an Intel-specific device that also just so happens to be conventional PCI, which means that this will force libvirt to add a `pcie-to-pci-bridge` controller. Not ideal.

I think `virtio-sound` might be a better default for non-x86 architectures, including existing ones. However, it doesn't look like support for it is currently implemented in libvirt, which clearly makes it a non-starter...

If `ich6` actually works on loongarch64 and friends (does it?) I guess it's as good an interim solution as any. Or would it be better to have no sound at all? Probably not.

andreabolognani:
Oh and we're missing at least `<rng>` and `<memballoon>`. There's an allowlist deciding whether those are to be added by default, and it needs to be extended to include loongarch64 guests.

tests/data/capabilities/kvm-loongarch64-domcaps.xml
lixianglai: Ok, I did take the contents of this file from the libvirt test case, where does it normally come from? Is the content of the domainCapabilities I get with the virt-manager --no-fork --debug command more appropriate?

tests/data/cli/compare/virt-install-kvm-loongarch64-basic.xml
lixianglai: Yes,After I add EFI information to domcapabilities, re-execute the command “pytest --regenerate-output”,the issue will be resolved.

lixianglai:
> Oh and we're missing at least `<rng>` and `<memballoon>`. There's an allowlist deciding whether those are to be added by default, and it needs to be extended to include loongarch64 guests.

Thanks,I'll add them in the guest.

tests/data/cli/compare/virt-install-kvm-loongarch64-basic.xml
lixianglai: loongarch64 uses ich6 to have sound, but the sound may not be continuous due to insufficient performance. I see other non-x86 architectures use ich9(The sound effects of ich9 and ich6 are the same),  For the next push, I will change the sound device to ich9,We can modify this together when libvirt supports virtio-sound.

virtinst/guest.py
```diff
@@ -990,7 +992,11 @@ def _add_default_usb_controller(self):
             # For pseries, we always assume OS supports usb3
             if qemu_usb3:
                 usb3 = True
-
+        elif self.os.is_loongarch64():
+            # For machvirt, we always assume OS supports usb3
+            if (qemu_usb3 and
+                self.conn.support.conn_machvirt_pci_default()):
```
andreabolognani, line 998: `conn_machvirt_pci_default()` is used to check whether the current version of libvirt is one where the `aarch64/virt` machine type defaults to PCI instead of the previous default of virtio-mmio. It's not relevant to loongarch64, you just need to check `qemu_usb3` as in the pSeries branch right above. Please update the comment as well.

virtinst/guest.py
```diff
@@ -1108,6 +1117,8 @@ def defaults_to_pcie(self):
             return True
         if self.os.is_riscv_virt():
             return True
+        if self.os.is_loongarch64():
+            return True
```
andreabolognani, line 1121: This was a very important spot that was missed the previous time around! Excellent catch :)

Things overall look good to me, but I recommend going one more time over all the spots where `self.os.is_*()` functions are used to make sure that nothing else is missed.

tests/data/capabilities/kvm-loongarch64.xml
```diff
@@ -0,0 +1,69 @@
...
+    </secmodel>
+  </host>
+
+  <guest>
+    <os_type>hvm</os_type>
+    <arch name='loongarch64'>
+      <wordsize>64</wordsize>
+      <emulator>/usr/bin/qemu-system-loongarch64</emulator>
+      <machine maxCpus='256'>virt</machine>
+      <domain type='qemu'/>
```
andreabolognani, line 58: With the latest update, this lost the `<domain type='kvm'/>` bit. This is reflected on the generated guest XML, where `<domain type="kvm">` was replaced with `<domain type="qemu">`. Did you generate it on a non-loongarch64 machine this time around?

Personally I don't particularly care whether KVM support is advertised or not, as I foresee that most people will be running loongarch64 VMs on x86_64 hardware for some time. But, for the sake of consistency, you should either generate this again so that KVM support is advertised, or update the names of the files so that they start with "qemu" rather than "kvm".

virtinst/guest.py
lixianglai: Thanks!I'll check back.

tests/data/capabilities/kvm-loongarch64.xml
lixianglai: ok

andreabolognani:
@lixianglai no more comments from my side, everything looks good. Of course we want to get loongarch64 support merged into libvirt first.

lixianglai:
> @lixianglai no more comments from my side, everything looks good. Of course we want to get loongarch64 support merged into libvirt first.

Thank you very much！  I will be pushing a new version of libvirt patch this week.

lixianglai:
@crobinso   Sorry to bother you, does this PR meet the entry requirements? Do I need to do anything?

tests/test_cli.py
```diff
@@ -1198,6 +1198,14 @@ def add_compare(self, cat, args, compbase, **kwargs):
 
 
 
+#####################
+# loongarch64 tests #
+#####################
+
+c.add_compare("--osinfo fedora19 --nodisks --pxe --connect " + utils.URIs.kvm_loongarch64, "kvm-loongarch64-basic")
```
andreabolognani, line 1205: @lixianglai I just noticed this but I think instead of using `utils.URIs.kvm_loongarch64` directly here you should define `URI-KVM-LOONGARCH64` like other architectures.

In terms of test coverage, the choice of having PXE only is a bit peculiar, and the use of `--nodisks` in particular means that we don't see whether the choice of disk model/controller would be the correct one. I would just copy and adapt the riscv64 tests, which ensure that both the headless and graphics scenarios are working as intended. Just use SPICE instead of VNC to ensure that sound is present.

andreabolognani: @lixianglai any chance you could take a look at this? A new virt-manager release might happen relatively soon, and it would be a shame if loongarch64 support couldn't make it in. Thanks!

lixianglai: I am very sorry for the late reply. Because I am busy with other work recently, I did not pay attention to the email message. I will modify it according to the opinion within this week, thank you very much!

lixianglai: Hi Andrea, I updated the loongarch64 test case with reference to the riscv64 test case on the latest code

lixianglai:
@andreabolognani @crobinso 
Hi Andrea and crobinso:
In this submission, I added a new patch, which adds optional bios type firmware for loongarch64 architecture, which essentially uses the -bios parameter to load QEMU_EFI.fd to achieve the purpose of using virt-manager internal snapshot.

I am not sure if this is within the norm, so I look forward to your comments and responses.
virt-manager-bios

andreabolognani:
@lixianglai thanks for updating the PR!

> In this submission, I added a new patch, which adds optional bios type firmware for loongarch64 architecture, which essentially uses the -bios parameter to load QEMU_EFI.fd to achieve the purpose of using virt-manager internal snapshot.

This doesn't look right at all.

While it's true that the edk2 binary can be loaded using QEMU's `-bios` option, the [firmware descriptor specification](https://gitlab.com/qemu-project/qemu/-/blob/master/docs/interop/firmware.json) very clearly states that the "bios" interface type is intended for "Traditional x86 BIOS interface. For example, firmware built from the SeaBIOS project", and that's the meaning both libvirt and virt-manager assign to it.

It's unfortunate that virt-manger currently doesn't support internal snapshots of running VMs when using UEFI, but hacking around this limitation by having incorrect information presented to the user and passed to libvirt is not an acceptable way to address that. Please drop the new commit.

Everything else looks good to me.

Related to the overall effort, when you have a minute can you please weight in on [this thread](https://lists.libvirt.org/archives/list/devel@lists.libvirt.org/thread/YZMV33UPKZRRQBN4XCCAW25IRV46LX57/) with information about support for TPM on loongarch64? That'd be very helpful. Thanks in advance!

lixianglai:
> @lixianglai thanks for updating the PR!
> 
> > In this submission, I added a new patch, which adds optional bios type firmware for loongarch64 architecture, which essentially uses the -bios parameter to load QEMU_EFI.fd to achieve the purpose of using virt-manager internal snapshot.
> 
> This doesn't look right at all.
> 
> While it's true that the edk2 binary can be loaded using QEMU's `-bios` option, the [firmware descriptor specification](https://gitlab.com/qemu-project/qemu/-/blob/master/docs/interop/firmware.json) very clearly states that the "bios" interface type is intended for "Traditional x86 BIOS interface. For example, firmware built from the SeaBIOS project", and that's the meaning both libvirt and virt-manager assign to it.
> 
> It's unfortunate that virt-manger currently doesn't support internal snapshots of running VMs when using UEFI, but hacking around this limitation by having incorrect information presented to the user and passed to libvirt is not an acceptable way to address that. Please drop the new commit.

Oh, that's a shame. Does the community have any plans or ideas to solve the problem of not being able to create snapshots when using UEFI firmware? Since the variable storage is configured read-only when flashmode is stateless, there is no write security problem in this mode. Can we allow the creation of internal snapshots in this mode？
> 
> Everything else looks good to me.
> 
> Related to the overall effort, when you have a minute can you please weight in on [this thread](https://lists.libvirt.org/archives/list/devel@lists.libvirt.org/thread/YZMV33UPKZRRQBN4XCCAW25IRV46LX57/) with information about support for TPM on loongarch64? That'd be very helpful. Thanks in advance!

Ok，I will reply to it as soon as possible。

phrdina:
Just a note about snapshots to clarify current state.

When UEFI is used the issue with internal snapshots for running VM is that QEMU will try to store memory state in the varstore as that is the first storage where it can be saved. That's the reason why libvirt and because of that virt-manager don't support creating internal snapshot of running VM.

Upstream virt-manager (not part of any release yet) supports creating external snapshots that work with running VM with UEFI and that's what upstream virt-manager will prefer now if supported by libvirt.

andreabolognani:
@phrdina thanks for providing this update! It's indeed amazing news. I've just tried it with my loongarch64 VM and it worked flawlessly :)

Do you think you could merge this PR now? AFAICT all of @crobinso's original feedback has been addressed, so I don't think it's necessary to wait for him to look at it again if me and you both are happy with the current version.

phrdina:
Will look at it today and if I don't see anything will merge it as well.

andreabolognani:
Thanks @phrdina! And thanks @lixianglai for your patience and excellent work both here and on the libvirt side :)

lixianglai:
@andreabolognani @phrdina @crobinso Thank you very much for your help in reviewing the code and giving your opinions, and also thank you for your recognition of my work !!!

PR #631, opened 2024-03-01
setup.py: move global args to install args

Paullgk:
Hello,
This PR is part of the current contribution of virt-manager packaging into Yocto `meta-virtualization` layer. This patch is used to have the ability to build and install virt-manager on non GUI system, and so, to avoid building the unneeded dependency in this use case. See the contribution mail submission here: https://lists.yoctoproject.org/g/meta-virtualization/message/8601

Waiting for your review, feel free to ask any questions or remarks.
Regards,
Paul Le Guen de Kerneizon, Savoir-faire Linux

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -242,6 +242,16 @@ class my_egg_info(setuptools.command.install_egg_info.install_egg_info):
 
 
 class my_install(setuptools.command.install.install):
+    setuptools.command.install.install.user_options += [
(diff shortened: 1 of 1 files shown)


phrdina:
Hi, thanks for your contribution but we are going to switch to meson so I'll close this PR as it will no longer be required. With meson this would be fixed.

PR #635, opened 2024-03-19
virt-xml: make it easier to switch between Spice and VNC graphics

phrdina:
When switching between Spice and VNC graphics it is usually not enough to just change the graphics type. For example spice uses additional devices to offer full functionality and they are added by virt-install/virt-manager whenever spice graphics is used.

With these changes using:

  `virt-xml {VM} --edit --graphics vnc`

will remove all Spice only related devices as well. Previously this invocation would fail as libvirt doesn't allow having other Spice devices without Spice graphics.

There might be use cases where video device needs to be changed together with graphics or in general any other devices changed together. To make that happen we now allow multiple edits at the same time. So for example:

  `virt-xml {VM} --edit --graphics vnc --edit --video virtio`

will now work.

### DIFF
--- a/virtinst/devices/video.py
+++ b/virtinst/devices/video.py
@@ -11,14 +11,24 @@ from ..xmlbuilder import XMLProperty
 class DeviceVideo(Device):
     XML_NAME = "video"
 
-    _XML_PROP_ORDER = ["model", "vram", "heads", "vgamem"]
-    model = XMLProperty("./model/@type")
+    _XML_PROP_ORDER = ["_model", "vram", "heads", "vgamem"]
     vram = XMLProperty("./model/@vram", is_int=True)
     vram64 = XMLProperty("./model/@vram64", is_int=True)
     ram = XMLProperty("./model/@ram", is_int=True)
     heads = XMLProperty("./model/@heads", is_int=True)
     vgamem = XMLProperty("./model/@vgamem", is_int=True)
     accel3d = XMLProperty("./model/acceleration/@accel3d", is_yesno=True)
+
+    def _set_model(self, val):
+        self._model = val
+        if self._model != "qxl":
+            self.ram = None
+            self.vgamem = None
+            self.vram64 = None
+    def _get_model(self):
+        return self._model
+    _model = XMLProperty("./model/@type")
+    model = property(_get_model, _set_model)
 
 
     ##################
--- a/virtinst/guest.py
+++ b/virtinst/guest.py
@@ -785,6 +785,13 @@ class Guest(XMLBuilder):
                 self.vcpus = defCPUs
         self.cpu.set_topology_defaults(self.vcpus)
 
+    def change_graphics(self, val, inst):
+        inst.type = val
+        if val == "spice":
+            self._add_spice_devices()
+        else:
+            self._remove_spice_devices(inst)
+
     def set_defaults(self, _guest):
         self.set_capabilities_defaults()
 
--- a/virtinst/guest.py
+++ b/virtinst/guest.py
@@ -504,6 +504,7 @@ class Guest(XMLBuilder):
     def remove_device(self, dev):
         self.devices.remove_child(dev)
         self._remove_duplicate_console(dev)
+        self._remove_spice_devices(dev)
 
     devices = XMLChildProperty(_DomainDevices, is_single=True)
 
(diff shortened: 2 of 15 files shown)


virtinst/guest.py
```diff
@@ -1206,3 +1207,26 @@ def _remove_duplicate_console(self, dev):
...
+            if channel.type == DeviceChannel.TYPE_SPICEVMC:
+                self.devices.remove_child(channel)
+
+    def _remove_spice_usbredir(self):
+        for redirdev in self.devices.redirdev:
+            if redirdev.type == "spicevmc":
+                self.devices.remove_child(redirdev)
+
+    def _remove_spice_devices(self):
+        if self.has_spice():
```
crobinso, line 1234: This makes me a little paranoid. Can you confirm that there's zero valid config in libvirt that allows spice audio, channel, redirdev if spice graphics isn't configured?

This could also give kinda unexpected results. Say you are using `cat foo.xml | virt-xml --edit --disk path=/foo` but the XML config has incorrect spice devices (maybe the input XML was hand edited). Suddenly the XML diff will show removal of the spice devices too. That isn't wrong, we may be fixing the XML, but it's unexpected IMO

I think just to be really safe here, and future proof incase somehow libvirt validation relaxes in the future, we do something like

```python
def _remove_spice_devices(self, rmdev):
    if (rmdev.DEVICE_TYPE != 'graphics' or
        rmdev.type != 'spice'):
        return
    ...
```

Or maybe we lift that check into `remove_device`, since this function is also reused for `change_graphics`

virtinst/guest.py
```diff
@@ -785,6 +785,13 @@ def sync_vcpus_topology(self, defCPUs):
                 self.vcpus = defCPUs
         self.cpu.set_topology_defaults(self.vcpus)
 
+    def change_graphics(self, val, inst):
+        inst.type = val
+        if val == "spice":
+            self._add_spice_devices()
+        else:
+            self._remove_spice_devices()
```
crobinso, line 793: coverage points out we don't hit this line. so we need a cli test case for switching spice -> vnc. Preferrably as a prep commit, so the diff here also demonstrates the XML bits changed. maybe test-for-virtxml isn't sufficient for input and we need a designated domain for it

virtinst/devices/video.py
```diff
@@ -11,15 +11,25 @@
...
     vram64 = XMLProperty("./model/@vram64", is_int=True)
     ram = XMLProperty("./model/@ram", is_int=True)
     heads = XMLProperty("./model/@heads", is_int=True)
     vgamem = XMLProperty("./model/@vgamem", is_int=True)
     accel3d = XMLProperty("./model/acceleration/@accel3d", is_yesno=True)
 
+    def _set_model(self, val):
+        self._model = val
+        if self._model != "qxl":
+            self.ram = None
```
crobinso, line 25: would be nice if this was demonstrated in the test suite too. something to fold into the spice -> vnc testing

phrdina:
> Thanks, mostly looks good, some specific comments inline.
> 
> I think last patch should be a separate PR though. It does more than listed, it also lets multiple `--add-device` and `--remove-device` calls work. This is probably a good thing for consistency, but we should test it, document it, and think about implications. Ideally the end case would be ability to do
> 
> ```
> $ virt-xml MYVM \
>     --edit 1 --disk type=foo \
>     --remove-device --network type=bar \
>     --add-device --sound model=wibble \
>     ...
> ```

Currently the patch allows only multiple `--edit` options, the example you've posted will fail with this error:

`ERROR    Conflicting options --edit, --add-device, --remove-device`

I wanted to keep the change as small as possible because this code might get backported.

Otherwise implementing the example above would definitely result in separate PR as it would be more changes.

> And any mix and matching.
> 
> The main issue I never wrapped my head around is what are we supposed to do if hotplug operations are requested but one of the hotplugs fails. We probably need to think about how error handling is done. Though right now virt-xml _does_ handle multiple devices sometimes, for things like `--controller usb2` IIRC, so we already have pre-existing behavior here

virtinst/guest.py
phrdina: Yeah we should test it. Will do and include video change as well into that test case so it demonstrate the real world usage.

crobinso:
Thanks for your patience and sorry for the delay, I've merged this now. But there's issues we need to straighten out before backporting patches anywhere.

> 
> Currently the patch allows only multiple `--edit` options, the example you've posted will fail with this error:
> 
> `ERROR Conflicting options --edit, --add-device, --remove-device`

Argh my bad, I should have tested the example. There is a behavior change though.

Previously, this worked (though I didn't recall that it does)

```console
$ ./virt-xml test --connect test:///default --print-diff --add-device --sound wibble --sound wobble
--- Original XML
+++ Altered XML
@@ -27,5 +27,7 @@
     <memballoon model="virtio">
       <address type="pci" domain="0x0000" bus="0x00" slot="0x02" function="0x0"/>
     </memballoon>
+    <sound model="wibble"/>
+    <sound model="wobble"/>
   </devices>
 </domain>
```

However this would fail:
```
$ ./virt-xml test --connect test:///default --print-diff --add-device --sound wibble --video wobble
ERROR    Only one change operation may be specified (conflicting options ['--sound', '--video'])
```

But after your patch, that succeeds. So minimally we need to add test coverage, or straighten this out.

Also, your patch requires multiple `--edit` listings, which IMO makes sense and is cleanest. Compared
to current code which only needs one `--add-device` call. We need to figure out how to join those.

man page needs an audit too. The `DESCRIPTION` field is wrong for example.

I'm not exactly sure when the multiple `--add-device` handling was added exactly, it probably slipped in accidentally at some point.

I will look into this and send a PR this week or otherwise start a discussion.

PR #636, opened 2024-03-20
Use GtkFileChooserNative

### LINKED ISSUE #315: Convert to FileChooserNative, for xdg portal usage
When selecting iso image for cd drive, you see gtk dialogs. First you see the dialog with already known locations. There you can press Choose file, and you will see gtk file picker dialog.

With support of xdg desktop portals, it may show kde file dialogs. It is better for users, as there are user's bookmarks in kde. 

**Distro**: Arch Linux
Virt-manager version: 3.2.0

michaelweghorn:
Use [GtkFileChooserNative](https://docs.gtk.org/gtk3/class.FileChooserNative.html) instead of [GtkFileChooserDialog](https://docs.gtk.org/gtk3/class.FileChooserDialog.html) to integrate better with the platform (e.g. use the portal implementation when run with GTK_USE_PORTAL=1, resulting in the KDE Frameworks implementation being used when
xdg-desktop-portal-kde is in use). Quoting from the [GtkFileChooserDialog doc](https://docs.gtk.org/gtk3/class.FileChooserDialog.html):

> If you want to integrate well with the platform you should use the
> GtkFileChooserNative API, which will use a platform-specific dialog if
> available and fall back to GtkFileChooserDialog otherwise.

Also replace the use of [GTK_STOCK_CANCEL](https://docs.gtk.org/gtk3/const.STOCK_CANCEL.html) and [GTK_STOCK_OPEN](https://docs.gtk.org/gtk3/const.STOCK_OPEN.html) which were deprecated in GTK 3.10 with the corresponding labels as suggested in their doc.

Fixes #315

### DIFF
--- a/virtManager/error.py
+++ b/virtManager/error.py
@@ -234,7 +234,7 @@ class vmmErrorDialog(vmmGObject):
 
     def browse_local(self, dialog_name, start_folder=None,
                      _type=None, dialog_type=None,
-                     choose_button=None, default_name=None,
+                     choose_label=None, default_name=None,
                      confirm_overwrite=False):
         """
         Helper function for launching a filechooser
@@ -246,19 +246,11 @@ class vmmErrorDialog(vmmGObject):
         """
         if dialog_type is None:
             dialog_type = Gtk.FileChooserAction.OPEN
-        if choose_button is None:
-            choose_button = Gtk.STOCK_OPEN
 
-        buttons = (Gtk.STOCK_CANCEL,
-                   Gtk.ResponseType.CANCEL,
-                   choose_button,
-                   Gtk.ResponseType.ACCEPT)
-
-        fcdialog = Gtk.FileChooserDialog(title=dialog_name,
-                                    parent=self.get_parent(),
-                                    action=dialog_type,
-                                    buttons=buttons)
-        fcdialog.set_default_response(Gtk.ResponseType.ACCEPT)
+        fcdialog = Gtk.FileChooserNative.new(title=dialog_name,
+                                             parent=self.get_parent(),
+                                             action=dialog_type,
+                                             accept_label=choose_label)
 
         if default_name:
             fcdialog.set_current_name(default_name)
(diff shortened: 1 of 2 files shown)


ptoscano (changes requested):
Thanks for the patch, which seems a good improvement. Since this seems to fix #315, please amend the commit message adding eg
```
Fixes #315
```
at the bottom of it.

Regarding the change on `choose_button`: this does not seem correct, as `Gtk.STOCK_*` are icon names, and it is actually used as such:
https://github.com/virt-manager/virt-manager/blob/dadb9a0d6fbfc4b1f106e59bd3b0ff06a5bf253a/virtManager/vmwindow.py#L557-L564

maybe what could be done here as part of this work is to change the `choose_button` parameter into `choose_label`, and change the only usage to pass `_("_Save")` instead of the icon name. Also, it seems that Gtk.FileChooserDialog has defaults for `accept_label` and `cancel_label`, so I'd always pass `None` for `cancel_label`, and use `None` as default for `accept_label`.

[1] https://docs.gtk.org/gtk3/class.FileChooserNative.html#properties

michaelweghorn:
> Thanks for the patch, which seems a good improvement. Since this seems to fix #315, please amend the commit message adding eg
> 
> ```
> Fixes #315
> ```
> 
> at the bottom of it.

Done.

> Regarding the change on `choose_button`: this does not seem correct, as `Gtk.STOCK_*` are icon names, and it is actually used as such:
> 
> https://github.com/virt-manager/virt-manager/blob/dadb9a0d6fbfc4b1f106e59bd3b0ff06a5bf253a/virtManager/vmwindow.py#L557-L564
> 
> maybe what could be done here as part of this work is to change the `choose_button` parameter into `choose_label`, and change the only usage to pass `_("_Save")` instead of the icon name. Also, it seems that Gtk.FileChooserDialog has defaults for `accept_label` and `cancel_label`, so I'd always pass `None` for `cancel_label`, and use `None` as default for `accept_label`.

Good catch! Thanks for the suggestion! That sounds good, I've implemented it this way now.

michaelweghorn:
> (...), so I'd always pass `None` for `cancel_label`, (...)

For the record, that's the default, so I just dropped that param completely.

crobinso:
Thanks @michaelweghorn and @ptoscano !

PR #640, opened 2024-03-24
build: switch fully to setuptools when possible

ptoscano:
Newer versions of `setuptools` (62.4.0+) provide a `setuptools.command.build` module to replace `distutils.command.build`; hence, try to use it when available, falling back to `distutils` otherwise as already done.

Include the switch to `setup()` to ease the migration.


crobinso (approved):
LGTM

PR #651, opened 2024-04-07
ci: Add test-against-libvirt-git.yml

crobinso:
Periodic job to run the test suite against libvirt.git

This is currently failing, so I'll disable it when it hits git master. I have patches for libvirt.git but they haven't hit the mailing list yet.

@pinotree @phrdina thoughts?


phrdina:
Looks good to me and yeah I was thinking about adding something like this as well.

PR #657, opened 2024-04-23
fix isolated pep 517 build

branchvincent:
This allows users to install via a standard isolated [PEP 517](https://peps.python.org/pep-0517/) build

- Adds `pyproject.toml` to explicitly define build requirements (i.e. `setuptools` and also `docutils`, since setup.py needs `rst2man`)
- Removes custom (disabled) `egg_info` command (`setuptools` needs it to [extract build requirements](https://github.com/pypa/setuptools/blob/1ed759173983656734c3606e9c97a348895e5e0c/setuptools/build_meta.py#L291))
- ~Don't attempt to run `gtk-update-icon-cache` / `glib-compile-schemas` unless found~

Before:

```console
$ pip install git+https://github.com/virt-manager/virt-manager
# ...
  × Building wheel for virt-manager (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [7 lines of output]
      # ...
      Didn't find rst2man or rst2man.py
```

After:

```console
$ pip install git+https://github.com/branchvincent/virt-manager@pep517
# ...
Successfully installed virt-manager-4.1.0
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,7 +19,6 @@ import subprocess
 
 import setuptools
 import setuptools.command.install
-import setuptools.command.install_egg_info
 try:
     # Use the setuptools build command with setuptools >= 62.4.0
     import setuptools.command.build
@@ -231,14 +230,6 @@ from %(pkgname)s import %(filename)s
 
         self.run_command("build_i18n")
         super().run()
-
-
-class my_egg_info(setuptools.command.install_egg_info.install_egg_info):
-    """
-    Disable egg_info installation, seems pointless for a non-library
-    """
-    def run(self):
-        pass
 
 
 class my_install(setuptools.command.install.install):
@@ -537,7 +528,6 @@ setuptools.setup(
         'build_i18n': my_build_i18n,
 
         'install': my_install,
-        'install_egg_info': my_egg_info,
 
         'configure': configure,
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,3 @@
+[build-system]
+requires = ["setuptools", "docutils"]
+build-backend = "setuptools.build_meta"
--- a/virt-manager.spec
+++ b/virt-manager.spec
@@ -132,6 +132,7 @@ machine).
 
 %{_datadir}/%{name}/ui/*.ui
 %{_datadir}/%{name}/virtManager
+%{python3_sitelib}/virt_manager-*.egg-info/
 
 %{_datadir}/%{name}/icons
 %{_datadir}/icons/hicolor/*/apps/*


setup.py
```diff
@@ -267,12 +258,12 @@ def finalize_options(self):
     def run(self):
         super().run()
 
-        if not self.distribution.no_update_icon_cache:
+        if not self.distribution.no_update_icon_cache and shutil.which("gtk-update-icon-cache"):
```
hartwork, line 261: Auto-magic is bad for distributions in that it needs to be found to not produce bad results without knowing and still end up as a hard requirement; better letter this crash or fail with an error please rather than not doing anything, silently. Same for the fix below. Thank you!

hartwork (changes requested):
Hi @branchvincent, interesting approach. Please see one detail comment below:

setup.py
branchvincent: yea that's perfectly reasonable, I didn't know if this was a hard requirement or not

crobinso:
Thanks for the patch. It's an interesting idea but is it useful if we install the application but then it doesn't work? Not a rhetorical question, I'm interested in opinions.

virt-install and virt-manager hard require C libraries with gobject-introspection bindings, which aren't expressable as `pyproject.toml` deps. So, `pip install` may succeed but the applications may immediately fail at startup.

branchvincent:
> virt-install and virt-manager hard require C libraries with gobject-introspection bindings

Thanks for confirming, dropped the check for `gtk-update-icon-cache` and `glib-compile-schemas`

ptoscano:
Please adjust the RPM spec file to also pick the newly installed egg files:
```
Checking for unpackaged file(s): /usr/lib/rpm/check-files /builddir/build/BUILD/virt-manager-v4.1.0-build/BUILDROOT
error: Installed (but unpackaged) file(s) found:
   /usr/lib/python3.13/site-packages/virt_manager-4.1.0-py3.13.egg-info/PKG-INFO
   /usr/lib/python3.13/site-packages/virt_manager-4.1.0-py3.13.egg-info/SOURCES.txt
   /usr/lib/python3.13/site-packages/virt_manager-4.1.0-py3.13.egg-info/dependency_links.txt
   /usr/lib/python3.13/site-packages/virt_manager-4.1.0-py3.13.egg-info/top_level.txt
```

branchvincent:
done

phrdina:
Hi, not sure we are ready to make this change and allow builds using `python -m build` and installing using `python -m pip install`. 

Currently virt-manager is installed into `/usr/share/virt-manager` or `$prefix/share/virt-manager` and by default it is not in `$PYTHONPATH`, to make it work we create our own wrappers that are installed into `$prefix/bin` and they inject the install location into the path. This doesn't work with `python -m pip install` because in that case `setuptools.command.install.install.prefix` is not set at all and our current `setup.py` code depends on it.

IMHO to address all these issues we need to modernize our `setup.py` and most likely start installing virt-manager into the python standard lib path.

crobinso:
> Hi, not sure we are ready to make this change and allow builds using `python -m build` and installing using `python -m pip install`.
> 
> Currently virt-manager is installed into `/usr/share/virt-manager` or `$prefix/share/virt-manager` and by default it is not in `$PYTHONPATH`, to make it work we create our own wrappers that are installed into `$prefix/bin` and they inject the install location into the path. This doesn't work with `python -m pip install` because in that case `setuptools.command.install.install.prefix` is not set at all and our current `setup.py` code depends on it.
> 
> IMHO to address all these issues we need to modernize our `setup.py` and most likely start installing virt-manager into the python standard lib path.

ACK, all that sounds good to me

phrdina:
Thanks for your contribution but I'm closing this PR. We agreed to go with meson and as I've already mentioned these changes are not enough to make isolated pep 517 build working correctly.

PR #668, opened 2024-05-20
xmleditor: make gtksourceview optional

jonner:
The only thing that GtkSourceView gives us is syntax highlighting and
auto-indent. When this library is not available, we can still offer xml
editing with a plain textview with very little lost functionality.

### DIFF
--- a/virtManager/xmleditor.py
+++ b/virtManager/xmleditor.py
@@ -7,13 +7,24 @@ import gi
 from virtinst import log
 
 # We can use either gtksourceview3 or gtksourceview4
+have_gtksourceview = True
 try:
     gi.require_version("GtkSource", "4")
     log.debug("Using GtkSource 4")
 except ValueError:  # pragma: no cover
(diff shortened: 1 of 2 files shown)


jonner:
As far as I can tell, the test failures are unrelated to my patch because they seem to also occur on the main branch.

virtManager/xmleditor.py
```diff
@@ -7,13 +7,24 @@
...
     gi.require_version("GtkSource", "4")
     log.debug("Using GtkSource 4")
 except ValueError:  # pragma: no cover
-    gi.require_version("GtkSource", "3.0")
-    log.debug("Using GtkSource 3.0")
-from gi.repository import GtkSource
+    try:
+        gi.require_version("GtkSource", "3.0")
+        log.debug("Using GtkSource 3.0")
+    except:
```
phrdina, line 18: Here we should catch only "ValueError" as that is used to indicate if it's not available.

PR #670, opened 2024-05-27
Improve riscv64 support

andreabolognani:
Make most things work out of the box when importing existing and upcoming disk images. As a side-effect, make importing aarch64 cloud images smoother too.

### DIFF
--- a/virtinst/install/installer.py
+++ b/virtinst/install/installer.py
@@ -187,6 +187,13 @@ class Installer(object):
         self._unattended_install_cdrom_device = dev.target
         guest.add_device(dev)
 
+        # The CDROM we just added might be a SCSI one. If that's the case,
+        # libvirt will automatically add the necessary controller for us, but
+        # depending on the architecture its choice of model might not be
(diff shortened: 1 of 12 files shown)


virtinst/install/installer.py
```diff
@@ -396,6 +396,14 @@ def _prepare(self, guest, meter):
...
             self._prepare_cloudinit(guest, meter)
 
+        # The information needed to perform an unattended installation might
+        # be passed to the guest in the form of a CDROM, which can potentially
+        # be attached via SCSI. libvirt will automatically add the SCSI
+        # controller for us, but depending on the architecture its choice of
+        # model might not be usable by the guest OS. Try to add a virtio-scsi
+        # controller for the best chance of the CDROM actually being picked up
+        guest.add_virtioscsi_controller()
+
```
phrdina, line 406: This doesn't look like the correct place. The cdrom device is added in function `_add_unattended_install_cdrom_device()` so we can add the controller there as well.

andreabolognani: You're absolutely right. Changed it.

PR #674, opened 2024-06-12
Add support for missing Hyper-V features and enable most of them by default

### LINKED ISSUE #154: RFE: enable more Hyper-V enlightenments by default
Originally filed here: https://bugzilla.redhat.com/show_bug.cgi?id=1851244

virt-install currently enables hv-time,hv-relaxed,hv-vapic,hv-spinlocks=0x1fff for guests which
support Hyper-V. There are several additional Hyper-V enlightenments supported by
QEMU and libvirt (see https://github.com/qemu/qemu/blob/master/docs/hyperv.txt and
https://libvirt.org/formatdomain.html#elementsFeatures) which could be useful to enable.
In particular, hv_stimer significantly reduces CPU usage when supporting
VMs are paused (see https://lore.kernel.org/kvm/20200625201046.GA179502@kevinolos/).

phrdina:
This patch series doesn't add support for `vendor_id` feature for now, that will be done in future.

We will not enable `reset` as it is not exported by modern Hyper-V versions and `reenlightenment` is mostly useful when migrating VMs with Hyper-V running inside and requires some other bits to make it work correctly.

Fixes #154

### DIFF
--- a/virtinst/domain/features.py
+++ b/virtinst/domain/features.py
@@ -31,6 +31,7 @@ class DomainFeatures(XMLBuilder):
     hyperv_spinlocks = XMLProperty("./hyperv/spinlocks/@state", is_onoff=True)
     hyperv_spinlocks_retries = XMLProperty("./hyperv/spinlocks/@retries",
                                            is_int=True)
+    hyperv_vpindex = XMLProperty("./hyperv/vpindex/@state", is_onoff=True)
     hyperv_synic = XMLProperty("./hyperv/synic/@state", is_onoff=True)
     hyperv_reset = XMLProperty("./hyperv/reset/@state", is_onoff=True)
 
--- a/virtinst/domain/features.py
+++ b/virtinst/domain/features.py
@@ -42,6 +42,7 @@ class DomainFeatures(XMLBuilder):
     hyperv_tlbflush = XMLProperty("./hyperv/tlbflush/@state", is_onoff=True)
     hyperv_ipi = XMLProperty("./hyperv/ipi/@state", is_onoff=True)
     hyperv_evmcs = XMLProperty("./hyperv/evmcs/@state", is_onoff=True)
+    hyperv_avic = XMLProperty("./hyperv/avic/@state", is_onoff=True)
 
     vmport = XMLProperty("./vmport/@state", is_onoff=True)
     kvm_hidden = XMLProperty("./kvm/hidden/@state", is_onoff=True)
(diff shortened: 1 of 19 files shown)


crobinso:
Nice work! ACK to everything up to the `support:` patch, feel free to push that now.

* One minor opinion about commit titles: for anything that extends the cli, I like a title like `cli: add --features hyperv.FOO=wibble` which is more interesting for reviewers and changelog readers than the internal field name. Also makes it easier for writing NEWS.md from changelog too. Something to consider for the future.

I'll add more in a follow up comment

virtinst/domain/features.py
```diff
@@ -49,6 +59,53 @@ class DomainFeatures(XMLBuilder):
...
+        _enable("ipi", requires=["vpindex"])
+
+        capsinfo = guest.lookup_capsinfo()
+
+        if guest.conn.caps.host.cpu.vendor == "Intel":
+            _enable("evmcs", requires=["vapic"])
+
+        if self.apic is True:
+            _enable("avic")
+
```
crobinso, line 108: * There's some minor pylint complaints around here:
```
virtinst/domain/features.py:68:8: W0102: Dangerous default value [] as argument (dangerous-default-value)
virtinst/domain/features.py:101:8: W0612: Unused variable 'capsinfo' (unused-variable)
```

crobinso:
@vittyvk I could use your help. phrdina is adding support to virt-install to enable many more hyper-v features by default for windows VMs, but I have some questions about qemu+hyperv

* Is it safe to specify `qemu -cpu ...,hv-FOO=on` even if host kernel doesn't support it? qemu doesn't choke?

* Similarly is it safe to enable `qemu -cpu ...,hv-BRAND-NEW-FOO=on` even if the guest OS doesn't support it? Like for older EOL windows versions.

* Looks like libvirt uses `qom-list-properties max-x86_64-cpu` to determine what `hv-*` flags are available. Is that checking against host kernel in any way, or just informing us that the flag is available in qemu?

crobinso:
@vittyvk thanks for the info. Indeed looks like libvirt uses hv-passthrough for checking host support too, per Michals original commits: https://gitlab.com/libvirt/libvirt/-/commit/10f47848648941c21ca8816458b33a37eb3a6d34

@phrdina so IMO this means we should not add hardcoded support.py bits, and limit this to only when libvirt has domcapabilities support. That way we aren't enabling things where the host may not support them. Simplifies our code and test matrix too.

Also even though it may not cause problems, it still feels a little weird to me to be enabling every new flag on old EOL OS like windows 7. Something like win10+ lineage and win2k16+ lineage seems more sane. But that would take more work to implement, so no objections if you leave it as is.

phrdina:
@crobinso Thanks for the review. Yeah that sounds reasonable to enable features only based on reported capabilities. I'll make the necessary modifications and update this PR.

PR #521, opened 2023-05-09
guest: Add default graphics for aarch64 

charlesa:
This request adds support for a video and graphics device on aarch64 based on whether libvirt reports virtio is supported. In order to pass pytest, updates to the fedora compare XML needed to be made.

PR #683, opened 2024-07-02
Add default graphics for aarch64 and riscv with virt machine

phrdina:
This adds test case for s390x where we do not set graphics by default to have full test code coverage.

### DIFF
--- a/tests/data/cli/compare/virt-install-aarch64-cdrom.xml
+++ b/tests/data/cli/compare/virt-install-aarch64-cdrom.xml
@@ -63,6 +63,18 @@
       <source mode="bind"/>
       <target type="virtio" name="org.qemu.guest_agent.0"/>
     </channel>
+    <channel type="spicevmc">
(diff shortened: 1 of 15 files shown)


andreabolognani:
@phrdina seems fine overall.

Looking at #521 though, which this PR obsoletes, I see that some checks were added for the availability of the virtio-gpu device and I wonder if it wouldn't make sense to have something like that. I'm not concerned about QEMU/libvirt being too old to support the device, but rather it being compiled out... Other devices, e.g. TPM and RNG, seem to have checks of that nature already.

phrdina:
I did consider keeping the check but the video code is there since 2017 and not sure if anyone would use QEMU with virtio compiled out. Sure, to be super safe we should probably add that check but for all architectures and that should be separate PR.

andreabolognani:
Yeah, adding the check as a follow-up would be perfectly fine. No issue on my side with merging this as-is.

PR #687, opened 2024-07-08
fix: no console gets selected after shutdown in serial

Worldwidebrine:
bug reproduce:
1. have both spice and serial
2. launch
4. switch to serial
5. halt
6. switch to spice

expected:
Nothing happens because spice should have been selected.

actual:
1. error page shows up
2. launch vm but error page is still there
3. switch to serial but spice shows up

### DIFF
--- a/virtManager/details/console.py
+++ b/virtManager/details/console.py
@@ -286,6 +286,7 @@ class _ConsoleMenu(vmmGObject):
         items = graphics + [[None, None, None]] + serials
 
         last_item = None
+        none_selected = True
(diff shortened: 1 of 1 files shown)


crobinso:
Thank you for the report, and the patch. I can reproduce. But after looking at the code I think there's a better way to fix this, which I did so in git now.

The `serial` menu entries are now not disabled in the UI for offline VMs. This fixes some of the menu weirdness, and allows pre-selecting serial as the console before the VM starts.

Please give the git code a try and file issues if you are still hitting problems. Thank you!

Issue #452, opened 2022-11-01
--cloud-init clouduser-ssh-key produces broken users configuration


gclawes:
**Distro**: Ubuntu 22.04
**cloud-init version**: 4.0.0 (also tested on 4.1.0)

Passing `--cloud-init clouduser-ssh-key=/path/to/id_ed25519.pub` causes an invalid `users:` configuration in cloud-init (`[REDACTED` added by me:
```
#cloud-config
users:
  - ssh-authorized-keys: ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIBYcQftiMEl/36wRFzyuTWITIYzO8ZTP+BJIZq0ORjX/ root@localhost
```

According to the `cloud-init` documentation (https://cloudinit.readthedocs.io/en/latest/topics/modules.html#users-and-groups): 
> Each users dictionary item must contain either a name or snapuser key, otherwise it will be ignored.

Without the ability to specify the username, this entry is always ignored.  The correct way to specify the ssh key for the `default` user is `ssh_authorized_keys:` (https://cloudinit.readthedocs.io/en/latest/topics/modules.html#authorized-keys)

Debug log attached: [virt-install-test.log](https://github.com/virt-manager/virt-manager/files/9910334/virt-install-test.log)

gclawes:
This seems related to #307 and #308

PR #694, opened 2024-07-23
virt-install: fix --clouduser-ssh-key

ncopa:
Fix the cloud-init config file for --clouduser-ssh key.

Fixes commit 22478f307dd9 (virt-install: Add --cloud-init clouduser-ssh-key=)

Resolves: https://github.com/virt-manager/virt-manager/issues/452


phrdina:
Hi, thanks for posting this patch. Unfortunately it doesn't work as there are multiple issue.

The correct attribute is `ssh_authorized_keys` because the old attribute `ssh-authorized-keys` was deprecated and works only if used for specific user. For the default user only `ssh_authorized_keys` works.

The second issue is more complex and will happen only if both root and coulduser keys are specified. If that happens the following config is created:
```yaml
users:
  - name: root
    ssh_authorized_keys:
      - rsa-ssh ......
ssh_authorized_keys:
  - rsa-ssh ......
```
which makes the default user removed.

The correct configuration needs to be like this:
```yaml
users:
  - default
  - name: root
    ssh_authorized_keys:
      - rsa-ssh ......
ssh_authorized_keys:
  - rsa-ssh ......
```
to have the default user created as well.

ncopa:
Thanks for reviewing it. Both issues have been addressed.

ncopa:
Should we add the `- default` if only root ssh key is set, or should setting root ssh key delete the clouduser (eg exclude the `- default`)?

phrdina:
I would say we should keep the default user to have consistent behavior. Before this patch it was inconsistent:
```
  root-password-file     =>     default user created
  root-ssh-key           =>     default user missing
  clouduser-ssh-key      =>     default user created
  both ssh keys          =>     default user missing
```
after this change default user will be created all the time.

Users can specify `user-data` and provide more details they want to do with the VM including any user configuration. The options above are just a shortcuts to the basic configuration in order to be able to ssh/log into the VM.

phrdina:
Btw: you can run `pytest --regenerate-output` to help you with fixing tests, review the changes it does and if it's ok commit it. That should help you make the test here working again.

Issue #488, opened 2023-01-30
RFE: add virt-install `--network backend.type=passt`


randrrandr1:
Hello! According to [docs](https://github.com/libvirt/libvirt/blob/2fa92efe9b286ad064833cd2d8b907698e58e1cf/docs/formatdomain.rst#userspace-slirp-or-passt-connection) libvirt supports passt backend for userspace network. Do you plan to support it?

Also I could not find virt-install command option to add port forwarding

PR #697, opened 2024-08-02
virt-install: implement passt backend for user network interface

phrdina:
Add new `--network backend.type` attribute, libvirt currently supports only `passt`. If not network type is specified virt-install will pick `user` type as it is required for `passt` backend.

This also introduces support to specify portForward and specifying port range as well.

For more details see [1].

[1] <https://libvirt.org/formatdomain.html#userspace-slirp-or-passt-connection>

Fixes: https://github.com/virt-manager/virt-manager/issues/488

### DIFF
--- a/tests/data/cli/compare/virt-install-many-devices.xml
+++ b/tests/data/cli/compare/virt-install-many-devices.xml
@@ -596,6 +596,18 @@
       <mac address="00:11:22:33:44:55"/>
       <model type="vmxnet3"/>
     </interface>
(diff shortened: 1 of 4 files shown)


sbrivio-rh:
> Do you know, if slirp isn't available, will libvirt use passt automatically for `type="user"` ?

@crobinso yes, see: https://gitlab.com/libvirt/libvirt/-/commit/8f34fd0c4cb3549e5d834234a20425c380bb2511

PR #700, opened 2024-08-06
Fix testsuite with libvirt.git

crobinso:


ptoscano:
Nothing super critical: rather than making `grep` a regexp, what about adding a separate `regrep` (or whatever, I'm bad at naming) to explicitly do regexp matching?

Otherwise I don't see anything wrong.

crobinso:
> Nothing super critical: rather than making `grep` a regexp, what about adding a separate `regrep` (or whatever, I'm bad at naming) to explicitly do regexp matching?
> 

I started doing that at first, but then reconsidered once I started coding it, made some of the error checking kinda ugly. When I tried swapping for `re.search` and saw only one string match had issues, I figured that was better approach. Plus `grep` naming kinda implies regex anyways, that has confused me in the past (even though I wrote the initial code anyways!)

PR #702, opened 2024-08-20
Add more launch security properties needed for SEV-SNP

berrange:
This extends the --launchSecurity arg to have all the new properties needed by SEV-SNP.

NB, I've not been able to test this on a real SNP host, so just had unit test coverage here, which at least exercises libvirt XML parsing and config validation.


berrange:
Hmm, tests fail as the libvirt being tested does not support SEV-SNP. What's the right way to make unit tests skipped on older libvirt lacking specific features ?

phrdina:
> Hmm, tests fail as the libvirt being tested does not support SEV-SNP. What's the right way to make unit tests skipped on older libvirt lacking specific features ?

You can set one of `prerun_check`, `precompare_check` or `predefine_check` parameters for the `add_compare` function. The value can be libvirt version string or function.

tests/test_cli.py
```diff
@@ -1224,12 +1224,13 @@ def add_compare(self, cat, args, compbase, **kwargs):
...
...zcynsjy4vOCKVBYrIXtBRrwBXo1yjZ0DosnnakBHRSmGVG1tSSppUU5b9TOJgQR2h9HMRlFSRbVNEItm91j0EQLoOCXzBEsAVJdwig75HIEHuIZimGbyOBTWasbmJPQtfx1+ElN5yBbe+wMprubuXe76IkhaUDO0wMLeQmfOnLOpkJeC8rA0Qm6f6wW/kPWkb+r1+2TDeEflwRHNspV2LwGpX5ctU5ruU7bcM82eatMqfkL7e4tnvjYmYJHxo8RHDICOxUqrSdZJZ/SpSEY66IXXs75sjYPGH/R1av+KzX71EhDpt+vEr3aOXd66uZ5NwWJIOC7A7mUDV7ObkpgtWiHxwMbpI2fllWsha+rLNJR5EwXWlItHJL6B7oBeUsJs62v6ZLHeJN81DHaOJOGvBR+Mv/V+tqrItLAtPfPrthwWvK96KDHsMKdlx0423FuPEf7AiYkupYi1GyoKQvMDNE5bQFrsDtaKbYCJKfoJJxdyVdZmDbtNnUpAvWwbj7gmfZ85QB0aPUzftWdSLNLL+xshCfQlO6ROril3Ir/7ITXE/1LKRS9719Fl6keYRVdExNiDotimnYWiVPZ20wDiWlmslOY8luWiXnWTm0xIvzbS16s7B8BpNZgbZ5BUX+eai7mmdvALDfPGtXIYtk4sPR54kvLyamIY5ZtshvW6gO57lpg5QM5TPyoAq2okufwC4IqAscnOoARO1aKxKGA9h+1DdV0eYIXqRCoW2mPRgYNBeimgbvAWu+bGJoCkF1U6+1psHFDd5fpsQPpft+WWxPMdffEP4yaQoXRFadjEL5Cq7Ib0ZFoj9d6a5UV35wo3SxnyWEsVb0hOGM2R/Fr4spYp3ms67XXZOvKvn3oOCOFx/8ySX1ErvpMqISioQgMK+PB4qqrISAOop0jGvUlxBZwN+meSXnA7CGG1ZzudB5pHx1za+vyb+U971iozFQ0/0CTe3hggNXo6OfBT+aaz8xsmV3TaJW83+Lvhn0XWuWt0Ztn59WTqyyqDeRFP07Z4awbaOzChJIMJTeretlit6azPH0f7K5CXdy60hQksJAgpTyAix2VBD7rcna4p5xvrxqbakUh//WLGegceJnpA9p3OuF7PUrrd54vuA7mad6fKBw==", "x86_64-launch-security-sev-snp-full", prerun_check="10.5.0")
```
phrdina, line 1234: To make coverage happy we need to add another test case, this should make it happy:
`c.add_invalid("--machine pc --launchSecurity type=sev-snp,policy=0x24", grep="SEV launch security requires a Q35 UEFI machine", prerun_check="10.5.0")`

phrdina:
Otherwise looks good.

tests/test_cli.py
berrange: Added that now

berrange: I don't know why the coverage report is still complaining about these missing lines when the unit test now covers them.

phrdina: I was trying to figure it as well. We are using fedora:latest container for that job so it has libvirt 10.1.0 that means we will not run these tests so coverage will complain. I did run it locally with newer libvirt and coverage is OK. I'll merge it.

PR #708, opened 2024-08-28
virt-xml: revert multi `--edit` support

crobinso:
CC @phrdina

This PR reverts the support for multiple `--edit` uses with virt-xml.
Ex `--edit --graphics type=vnc --edit --video model=vga`

These were the last 2 commits in https://github.com/virt-manager/virt-manager/pull/635

Additionally some test cases are added for historical behavior we need to regression test.

Playing with this some more I found some deeper problems. For example

```console
$ ./virt-xml test-for-virtxml --connect test://`pwd`/tests/data/testdriver/testsuite.xml --print-diff --edit 1 --video model=FOO --edit 2 --sound model=BAR
--- Original XML
+++ Altered XML
@@ -180,14 +180,14 @@
     <graphics type="vnc" port="-1" autoport="yes">
       <listen type="address"/>
     </graphics>
-    <sound model="sb16"/>
+    <sound model="BAR"/>
     <sound model="es1370"/>
     <sound model="ich6"/>
     <video>
       <model type="vmvga" vram="16384" heads="1" primary="yes"/>
     </video>
     <video>
-      <model type="cirrus" vram="16384" heads="3"/>
+      <model type="FOO" vram="16384" heads="3"/>
     </video>
     <hostdev mode="subsystem" type="usb" managed="yes">
       <source>
```

Compare the XML diff to the command line, you'll see that `--edit 1` is actually paired with `--sound model=BAR`, which is not how things are ordered on the command line.

The order the device options are applied is dictated by `cli.VIRT_PARSERS`, not the order they are specified by the user. This seems like it should be a simple fix, but it isn't: argparse really isn't designed to care about the order of different options on the command line, so it's hard to determine that `--sound model=BAR` is actually the second listed XML option.

There's other weirdness too, though it's mostly strange plays on previous weird behavior

* `--edit --video model=vga --edit --sound model=ich9 --video model=qxl` works
* `--add-device --sound model=foo --video model=bar` was previously rejected but now works
* `--remove-device --video model=vmvga --sound model=sb16` was previously rejected by now works

So I think we need to revert and go back to the drawing board. I will file an issue with some more thoughts


phrdina:
That's a nice catch, didn't realize the ordering is dictated by `cli.VIRT_PARSERS`. Agreed, this is a major issue and we need to figure out different approach.

crobinso:
I filed this issue. We can probably add some custom argparse behavior to give us what we need, but it would need testing: https://github.com/virt-manager/virt-manager/issues/709

phrdina:
Sounds good. I'll look into it.

PR #721, opened 2024-09-11
domain: Use KEEP_TPM flag for domain rename

crobinso:
If the VM has implicit TPM state, use the VIR_DOMAIN_UNDEFINE_KEEP_TPM flag to preserve that state when renaming the VM (if libvirt is new enough).

The state is stored based on VM UUID and nothing else, and the UUID is preserved during rename, so we don't need to do any of the same trickery that's required for nvram duplication.

### DIFF
--- a/virtManager/object/domain.py
+++ b/virtManager/object/domain.py
@@ -436,6 +436,10 @@ class vmmDomain(vmmLibvirtObject):
         return bool(self.get_xmlobj().is_uefi() or
                     self.get_xmlobj().os.nvram)
 
+    def has_tpm_state(self):
+        return any(tpm.type == "emulator"
+                   for tpm in self.get_xmlobj().devices.tpm)
+
     def is_persistent(self):
         return bool(self._backend.isPersistent())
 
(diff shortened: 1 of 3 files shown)


virtManager/object/domain.py
```diff
@@ -436,6 +436,11 @@ def has_nvram(self):
         return bool(self.get_xmlobj().is_uefi() or
                     self.get_xmlobj().os.nvram)
 
+    def has_tpm_state(self):
+        for tpm in self.get_xmlobj().devices.tpm:
+            if tpm.type == "emulator":
+                return True
```
ptoscano, line 442: please explicitly return `False` at the end of the function, otherwise it returns `None`

another possibility is to use `any()` -- something like:
```python
return any(tpm.type == "emulator" tpm in self.get_xmlobj().devices.tpm)
```

ptoscano (changes requested):

virtManager/object/domain.py
crobinso: I'll use the oneliner, thanks. My primary complaint with oneliners is they obscure code coverage testing but it doesn't matter in this instance

PR #727, opened 2024-09-18
virt-xml: add `--convert-to-vnc`

crobinso:
This series adds a `virt-xml --convert-to-vnc` option, for handling everything I can think of to try and convert an existing VM to use VNC.

The primary usecase is spice->VNC, with particular emphasis on the case where spice is no longer available for the distro, like certain RHEL versions. So we strip out all spice-dependent devices like `qxl`, `spicevmc` bits, etc. Cockpit is already opencoding a limited version of this, but they would prefer if virt-xml handles it. Their code is here: https://github.com/cockpit-project/cockpit-machines/blob/main/src/libvirt-xml-update.js#L309

This also establishes a new pattern with `--convert-to-X` style compound operations for virt-xml, which we can follow for future `--convert-to-q35` for example. We could add suboptions if users need to opt out of certain behaviors of the command. There's one mostly demo option added, `qemu-vdagent=on|off` which defaults to `off`.

### DIFF
--- a/virtinst/guest.py
+++ b/virtinst/guest.py
@@ -828,6 +828,84 @@ class Guest(XMLBuilder):
 
         self.add_q35_pcie_controllers()
 
+    def _convert_spice_gl_to_egl_headless(self):
+        if not self.has_spice():
+            return
+
+        spicedev = [g for g in self.devices.graphics if g.type == "spice"][0]
+        if not spicedev.gl:
+            return
+
+        dev = DeviceGraphics(self.conn)
+        dev.type = "egl-headless"
+        dev.set_defaults(self.conn)
+        if spicedev.rendernode:
+            dev.rendernode = spicedev.rendernode
+        self.add_device(dev)
+
+    def _convert_to_vnc_graphics(self):
+        """
+        If there's already VNC graphics configured, we leave it intact,
+        but rip out all evidence of other graphics devices.
+
+        If there's other non-VNC, non-egl-headless configured, we try to
+        inplace convert the first device we encounter.
+
+        If there's no graphics configured, set up a default VNC config.`
+        """
+        vnc_devs = [g for g in self.devices.graphics if g.type == "vnc"]
+        # We ignore egl-headless, it's not a true graphical frontend
+        other_devs = [g for g in self.devices.graphics if
+                      g.type != "vnc" and g.type != "egl-headless"]
+
+        # Guest already had a vnc device.
+        # Remove all other devs and we are done
+        if vnc_devs:
+            for g in other_devs:
+                self.remove_device(g)
+            return
+
+        # We didn't find any non-vnc device to convert.
+        # Add a vnc device with default config
+        if not other_devs:
+            dev = DeviceGraphics(self.conn)
+            dev.type = dev.TYPE_VNC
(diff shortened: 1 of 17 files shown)


virtinst/guest.py
```diff
@@ -845,6 +845,37 @@ def _convert_to_vnc_graphics(self):
...
+        """
+        if not self.devices.video:
+            videodev = DeviceVideo(self.conn)
+            videodev.set_defaults(self)
+            self.add_device(videodev)
+            return
+
+        for video in self.devices.video:
+            # Disable accel3d which is largely tied to spice
+            video.accel3d = None
```
phrdina, line 867: Not sure about this. For VNC we can add `egl-headless` graphics type which makes virgl work with VNC.

virtinst/guest.py
```diff
@@ -845,6 +845,37 @@ def _convert_to_vnc_graphics(self):
...
+
+        qxl_devs = [v for v in self.devices.video if v.model == "qxl"]
+        other_devs = [v for v in self.devices.video if v not in qxl_devs]
+        if not qxl_devs:
+            return
+
+        dev = qxl_devs.pop(0)
+        for video in qxl_devs + other_devs:
+            # We remove other devs, not sure how to deal with mixed setup
+            self.remove_device(video)
```
phrdina, line 877: We should not remove all other devices if there are multiple video devices including `qxl`. We can have situation where VM has `bosch` as primary and `qxl` as another video device. This would remove both devices and replaced it with default video device (in most cases `virtio`).

I would change the logic to keep other video devices intact and modify only `qxl` devices. There can be multiple `qxl` devices so if it is `primary` we should replace it with default device, it it's not `primary` we can replace it with `virtio` or remove it completely.

crobinso: Nice, will fix

virtinst/guest.py
crobinso: Good call. I did not understand `egl-headless` until 30 mins ago :) Will fix

virtinst/guest.py
```diff
@@ -862,6 +862,22 @@ def _convert_to_vnc_graphics(self):
...
+        if not self.devices.video:
+            videodev = DeviceVideo(self.conn)
+            videodev.set_defaults(self)
+            self.add_device(videodev)
+            return
+
+        qxl_devs = [v for v in self.devices.video if v.model == "qxl"]
+        for dev in qxl_devs:
+            dev.clear()
+            dev.set_defaults(self)
```
phrdina, line 883: This will fail for some VMs, for example Windows 11 will get `bocsh`. Libvirt allows only `qxl` or `virtio` as secondary video device, any other type is rejected.

We need to hard-code `virtio` here or use
``` python
        if (guest.lookup_domcaps().supports_video_virtio() and
            guest.osinfo.supports_virtiogpu()):
```
to decide if we convert it to `virtio` or remove it.

crobinso: Another nice catch :) Ok I fixed it without the open coding check. Just call `dev.set_defaults()` and if we get back a non-virtio model for a non-primary device, remove the device.

phrdina (approved):
Everything looks good now.

PR #728, opened 2024-09-18
virt-xml: cleanup action handling in preparation for multi --edit support

crobinso:
This series cleans up a bunch of miscellaneous things with virt-xml option processing.

The main bit it adds a `Action` class, which contains all the data to describe a single XML operation like `--edit 1 --disk path=/foo` or `--add-device --network default`, etc. This does not add any new functionality, but it cleans up a lot of mess that would prevent adding proper multi action support in the future.

Commits that follow on from that are small bugfixes and code improvements in the same area.

### DIFF
--- a/virtinst/virtxml.py
+++ b/virtinst/virtxml.py
@@ -44,11 +44,11 @@ def get_diff(origxml, newxml):
     return diff
 
 
-def set_os_variant(options, guest):
-    if options.os_variant is None:
+def set_os_variant(guest, os_variant):
+    if os_variant is None:
         return
 
-    osdata = cli.parse_os_variant(options.os_variant)
+    osdata = cli.parse_os_variant(os_variant)
     if osdata.get_name():
         guest.set_os_name(osdata.get_name())
 
@@ -231,12 +231,13 @@ def _find_objects_to_edit(guest, action_name, editval, parserclass):
     return inst
 
 
-def action_edit(action, guest, options):
+def action_edit(action, guest):
     parserclass = action.parserclass
+    parservalue = action.parservalue
     selector = action.selector
 
     if parserclass is cli.ParserXML:
-        cli.parse_xmlcli(guest, options)
+        cli.parse_xmlcli(guest, parservalue)
         return []
 
     if parserclass.guest_propname:
@@ -252,22 +253,23 @@ def action_edit(action, guest, options):
 
     devs = []
     for editinst in xmlutil.listify(inst):
-        devs += cli.run_parser(options, guest, parserclass,
+        devs += cli.run_parser(guest, parserclass, parservalue,
                                editinst=editinst)
     return devs
 
 
-def action_add_device(action, guest, options, input_devs):
+def action_add_device(action, guest, os_variant, input_devs):
     parserclass = action.parserclass
+    parservalue = action.parservalue
 
-    set_os_variant(options, guest)
+    set_os_variant(guest, os_variant)
 
     if input_devs:
         for dev in input_devs:
             guest.add_device(dev)
         devs = input_devs
     else:
-        devs = cli.run_parser(options, guest, parserclass)
+        devs = cli.run_parser(guest, parserclass, parservalue)
         for dev in devs:
             dev.set_defaults(guest)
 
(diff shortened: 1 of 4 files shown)


virtinst/virtxml.py
```diff
@@ -63,37 +63,111 @@ def defined_xml_is_unchanged(conn, domain, original_xml):
...
+        self.selector = selector
+        # ParserDisk, etc
+        self.parserclass = parserclass
+        # ex for `--disk path=/foo` this is "path=/foo"
+        self.parservalue = parservalue
+
+        self.is_edit = self.action_string == "edit"
+        self.is_add_device = self.action_string == "add-device"
+        self.is_remove_device = self.action_string == "remove-device"
+        self.is_build_xml = self.action_string == "build-xml"
```
ptoscano, line 85: please make these as properties, rather than instance variables: this avoids the risk to modify them & to have a inconsistency between these and `action_string`

virtinst/virtxml.py
```diff
@@ -63,37 +63,111 @@ def defined_xml_is_unchanged(conn, domain, original_xml):
...
                "(conflicting options %s)") %
-               [c.cli_flag_name() for c in collisions])
+               [c[0].cli_flag_name() for c in collisions])
+
+    parserclass, parservalue = collisions[0]
+    return parserclass, parservalue
 
-    return collisions[0]
+
+def parse_actions(conn, options):
```
ptoscano, line 153: nitpick: since this function parses the command line options and returns only one Action, I'd name it `parse_action()` (singular); this makes the usage of it more readable:
```python
    action = parse_action(conn, options)
```

virtinst/virtxml.py
```diff
@@ -256,21 +340,23 @@ def start_domain_transient(conn, xmlobj, devs, action, confirm):
...
         msg_success = _("Device hotunplug successful.")
         msg_fail = _("Error attempting device hotunplug: %(error)s")
-    elif action == "update":
+    elif action.is_edit:
         msg_confirm = _("%(xml)s\n\nUpdate this device for the guest "
                         "'%(domain)s'?")
         msg_success = _("Device update successful.")
         msg_fail = _("Error attempting device update: %(error)s")
+    else:
+        raise xmlutil.DevError("update_changes without get_hotplug_name()")
```
ptoscano, line 355: even if this seems like a failsafe, the message seems odd to me

virtinst/virtxml.py
```diff
@@ -181,35 +263,37 @@ def action_add_device(guest, options, parserclass, devs):
...
     return devs
 
 
+def perform_action(action, guest, options, input_devs):
+    if action.is_add_device:
+        return action_add_device(action, guest, options, input_devs)
+    if action.is_remove_device:
+        return action_remove_device(action, guest)
+    assert action.is_edit
+    return action_edit(action, guest, options)
```
ptoscano: even if this is failsafe code, I'd use an approach similar to what done in `update_changes()`, ie
```python
if action.is_edit:
    return action_edit(action, guest, options)
raise xmlutil.DevError("perform_action(): unhandled action %s" % action.action_string)
```
(or `AssertionError` instead of `xmlutil.DevError`)

crobinso: all good points, I've addressed them now. thanks pino!

virtinst/virtxml.py
```diff
@@ -109,103 +228,82 @@ def _find_objects_to_edit(guest, action_name, editval, parserclass):
...
+def perform_action(action, guest, options, input_devs):
+    if action.is_add_device:
+        return action_add_device(action, guest, options.os_variant, input_devs)
+    if action.is_remove_device:
+        return action_remove_device(action, guest)
+    if action.is_edit:
+        return action_edit(action, guest)
+    raise xmlutil.DevError(
+        "perform_action() incorrectly called with action_name=%s" %
+        action.action_name)
```
ptoscano, line 304: `action_string`

virtinst/virtxml.py
```diff
@@ -252,21 +350,25 @@ def start_domain_transient(conn, xmlobj, devs, action, confirm):
...
-    elif action == "update":
+    elif action.is_edit:
         msg_confirm = _("%(xml)s\n\nUpdate this device for the guest "
                         "'%(domain)s'?")
         msg_success = _("Device update successful.")
         msg_fail = _("Error attempting device update: %(error)s")
+    else:
+        raise xmlutil.DevError(
+                "update_changes() incorrectly called with action=%s" %
+                action.action_name)
```
ptoscano, line 371: `action_string`

crobinso: argh, good catch. I'll fix by using `action_name` everywhere since there's already a mix of it and `action_string`

Issue #720, opened 2024-09-10
RFE: properly handle changing `q35` <-> `pc` in new VM 'Customize' wizard


crobinso:
Currently if the 'New VM' wizard will default to machine `q35`, and you change that to `pc` in the `Customize` wizard at the end of install, the VM fails to boot. This is https://github.com/virt-manager/virt-manager/issues/398 and will be fixed shortly, but the fix isn't going to be a full solution.

There's several default VM defaults that are determined by `q35` vs `pc`. https://github.com/virt-manager/virt-manager/issues/719 documents the inverse. The fix for issue #398 will solve the pcie vs pci issue, but not the other ones.

The architecture of the New VM wizard makes this difficult to fix in a future proof way. Once we enter the `Customize` wizard, all the initial defaults have been filled in, taking into account architecture, OS choice, install config, etc. The `Customize` wizard then only sees that fully filled in XML. But the bits that fill in q35 dependent defaults expect to run on the mostly empty XML, so that if it encounters say an already filled in disk `bus` value, it assumes the user explicitly requested and won't attempt to set any default.

We could use the `--convert-to-q35` functionality suggested in https://github.com/virt-manager/virt-manager/issues/719 . We would need some inverse `--convert-to-pc` function too though.

Another option is moving UI `Machine`, and possibly `Firmware` options to the `New VM` wizard itself, probably on the last page under an `Advanced` expander. But this comes with all the pain of moving UI, users will have to adjust etc.

There's also the option of doing some kind of bigger reorg of the handoff from `New VM` to `Customize` wizard, or the backing `Guest` instance, so these types of things can be more easily regenerated.

I don't really have my thoughts settled here. It may not even really be worth it for the `pc` case once https://github.com/virt-manager/virt-manager/issues/398 is fixed.

crobinso:
`--convert-to-q35` has been added. I think the plan here should be to add the inverse, `--convert-to-pc`, and then use the same infrastructure in the UI when changing machine type in the `Customize` wizard.

PR #729, opened 2024-09-23
virt-xml: add `--edit --convert-to-q35`

crobinso:
This PR is on top of https://github.com/virt-manager/virt-manager/pull/727 since they overlap

This adds `virt-xml --edit --convert-to-q35`. The XML bits we actually edit are layed out in https://github.com/virt-manager/virt-manager/issues/720 plus converting ich6 -> ich9 sound, and clearing disk `<address>` blocks.

In my limited testing, VMs handle this mostly OK:

* Fedora 39 reboots fine and everything works. Even manually converting virtio storage to `sata` config and Fedora doesn't mind.
* Win10 will fail to boot. After a couple reboots it drops you to the repair menu. If you boot once into safe mode, it seems to figure everything out, and subsequent reboots boot fine.
* Win11 rejects machine=pc in the first place, so we shouldn't need to worry about that.

### DIFF
--- a/man/virt-xml.rst
+++ b/man/virt-xml.rst
@@ -197,6 +197,31 @@ GUEST OS OPTIONS
     See virt-install(1) documentation for more details about ``--os-variant/--osinfo``
 
 
+CONVERSION OPTIONS
+==================
+
+``--convert-to-q35``
+^^^^^^^^^^^^^^^^^^^^
+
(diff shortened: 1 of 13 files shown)


crobinso:
I rebased this to be independent of `--convert-to-vnc`

phrdina (approved):
Looks good to me.

PR #730, opened 2024-09-24
snapshots: Don't save outdated screenshot data

crobinso:
Current code will take a screenshot when the 'New Snapshot' dialog is initially launched, and then use that cached content as the screenshot for the snapshot... which may be totally out of date by the time the user enters all the snapshot info.

Instead, drop the snapshot preview UI in the 'New Snapshot' wizard entirely, and use a screenshot that is fetched after the user actually clicks 'Finish'


phrdina:
Agreed, we probably don't need to display the screenshot in the `New Snapshot` UI.

Issue #497, opened 2023-03-04
`--cloud-init` + `--events on_reboot=restart` and the VM still fully shuts down after VM initiated reboot


lebfr:
**Distro**: Debian 11
**Version**:  3.2.0

Steps:
virt-install --events on_reboot=restart \
--name test2 --memory 2048 --vcpus 1 --cpu host \
--os-type=linux --os-variant debian10 \
--import --disk vol=default/test2.qcow2 \
--network network=lan,model=virtio \
--graphics vnc \
--cloud-init disable=on,meta-data=/vm/templates/test2-meta-data,user-data=/vm/templates/user-data \
--noautoconsole \
--autostart

Expected result:
virsh dumpxml test2 | grep on_reboot
  <on_reboot>restart</on_reboot>

Actual result:
virsh dumpxml test2 | grep on_reboot
  <on_reboot>destroy</on_reboot>

[This code](https://github.com/virt-manager/virt-manager/blob/33ff193ee9fcfdb74f95d946a1b93239a1a12a61/virtinst/install/installer.py#L237) is always executed. The virtual machine always shutdown on the first reboot. The --events option does not affect the behavior. It is not always comfortable when importing a virtual machine.

lebfr:
I changed the behavior for myself but my code looks like a very dirty hack.

lebfr:
https://github.com/lebfr/virt-manager/commit/6a386f3939dc3872c7ffe50330967cce1c747f03

kettlewell:
Thanks for the code... I can vouch that in my use case, it works as intended... I've been pulling my hair out trying to figure out why virt-install wasn't rebooting when cloud-init was setting the reboot mode in the power_state... 

@lebfr were you going to create a PR for that, or was that just a starting point for someone else to run with?

crobinso:
Sorry for late response. virt-install forcing shutdown is intended behavior. When you use `--cloud-init` we generate one time cloud-init media meant for initial VM configuration, and then we delete it. The only way to have the VM know that this is intended as one time media, is to force it to fully shut down, so we can change the VM config to remove the deleted media. New VM config can not be applied when VM does a warm reboot/reset.

Can yall explain what your usecase is? You want virt-install's generated cloud-init media to persist and retrigger for every boot? You want warm reboot for some usecase? Or is VM left in shutdown state but you want it running?

kettlewell:
For my use case, I want a final reboot after cloud-init is done, so I leverage the power_state config similar to this:
[...]

PR #734, opened 2024-10-09
virt-install: Make --cloud-init reboot by default (again)

crobinso:
Patch 1 comment is below. Patch 2 is a cleanup on top.

This basically reverts this commit from virt-manager 4.0.0

```
commit https://github.com/virt-manager/virt-manager/commit/825ec644b859b99d6f9f155b04ef3ac07eca1fc2
Author: Cole Robinson <crobinso@redhat.com>
Date:   Tue Mar 1 10:32:01 2022 -0500

    installer: Do not force reboot with --cloud-init

    Resolves: https://github.com/virt-manager/virt-manager/issues/189
```

After that commit, using `--cloud-init` or `--unattended` would
basically imply `--noreboot`

For some usage of `--cloud-init`, like in the initial report, this
makes sense to do. virt-install will drop you into the cloud image,
you mess without, run `poweroff`, and are surprised when virt-install
reboots the VM. I agreed, and made the change

But changing this was a bad idea.

cloud-init and unattended VMs have 2 XML phases. First XML is booting
off temporary media (the generated cloudinit iso), second XML is the
permanent boot config (booting off disk). We need to force the VM
to fully poweroff, so that the second XML config takes effect. We
make that happen with `<on_reboot>destroy</on_reboot>`, which tells
libvirt/qemu to poweroff the VM when it receives a guest reboot
notification. virt-install then defaults to restarting the VM when
it detects shutdown, to manually replicate the VM requested reboot,
but still get the second XML config to take effect.

The perennial problem with this is that, from outside the VM, we
don't know if the user inside the VM requested `poweroff` or `reboot`.
virt-install emulates the reboot behavior, and provides `--noreboot`
to override it. It's still confusing if a user puts a `poweroff`
command at the end of an anaconda kickstart file, and sees the VM
reboot, but it's been that way forever.

Except with that commit above, we flipped it for --cloud-init and
--unattended. Now, users who do `poweroff` are getting expected
behavior, but users who request `reboot` are not.

We could go further and add a `--yesreboot` option or similar. But
for consistency sake I think we should just revert that behavior,
and tell users to use --noreboot like usual.

Resolves: https://github.com/virt-manager/virt-manager/issues/497

### DIFF
--- a/virtManager/createvm.py
+++ b/virtManager/createvm.py
@@ -2031,7 +2031,7 @@ class vmmCreateVM(vmmGObjectUI):
             # Probably means guest had no 'install' phase, as in
             # for live cds. Try to restart the domain.
             vm.startup()  # pragma: no cover
-        elif installer.has_install_phase():
+        elif installer.requires_postboot_xml_changes():
             # Register a status listener, which will restart the
             # guest after the install has finished
             def cb():
--- a/virtinst/install/installer.py
+++ b/virtinst/install/installer.py
@@ -507,6 +507,8 @@ class Installer(object):
         into the guest. Things like LiveCDs, Import, or a manually specified
         bootorder do not have an install phase.
         """
+        if self.has_cloudinit() or self.has_unattended():
+            return True
         if self._no_install:
             return False
         return bool(self._cdrom or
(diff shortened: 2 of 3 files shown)


phrdina:
Agreed, keeping it consistent with other form of installation makes sense.

Issue #652, opened 2024-04-11
Release >=4.2.0 with meson rather than distutils/setuptools


hartwork:
Hi virt-manager team,

thanks for the recent work getting off dead distutils in 231a3dbc875333c73347fc6be8b39d9fc5542460. It would be great to have a release with the new Setuptools support, so that e.g. Gentoo can start using PEP 517 builds for it (https://bugs.gentoo.org/909876).

Thank you!

CC @ptoscano @crobinso

ptoscano:
> thanks for the recent work getting off dead distutils in [231a3db](https://github.com/virt-manager/virt-manager/commit/231a3dbc875333c73347fc6be8b39d9fc5542460). It would be great to have a release with the new Setuptools support, so that e.g. Gentoo can start using PEP 517 builds for it (https://bugs.gentoo.org/909876).

Unfortunately I don't do releases, that's mostly up to Cole.

If you are in pressing need (at least from what I can see from the gentoo bug), I recommend backporting 4f4e73d0170fdd7efdc552c7baebdf739b60df6a and 231a3dbc875333c73347fc6be8b39d9fc5542460 to gentoo until a new release is available. They are only build system changes, so they have no impact on the actual functionalities of virt-manager.

hartwork:
Hi @ptoscano, thanks for your reply!

> > thanks for the recent work getting off dead distutils in [231a3db](https://github.com/virt-manager/virt-manager/commit/231a3dbc875333c73347fc6be8b39d9fc5542460). It would be great to have a release with the new Setuptools support, so that e.g. Gentoo can start using PEP 517 builds for it (https://bugs.gentoo.org/909876).
> 
> Unfortunately I don't do releases, that's mostly up to Cole.

I see, @crobinso is there a chance for a new release in the coming weeks?

> If you are in pressing need (at least from what I can see from the gentoo bug), I recommend backporting [4f4e73d](https://github.com/virt-manager/virt-manager/commit/4f4e73d0170fdd7efdc552c7baebdf739b60df6a) and [231a3db](https://github.com/virt-manager/virt-manager/commit/231a3dbc875333c73347fc6be8b39d9fc5542460) to gentoo until a new release is available. They are only build system changes, so they have no impact on the actual functionalities of virt-manager.

[...]

PR #741, opened 2024-10-25
Update setup.py to allow PEP 517 build

phrdina:
This implements smallest possible changes to current setuptools build system in order allow PEP 517 builds.

Key changes with the new python build system:
- We cannot figure out prefix when building/installing virt-manager due to the fact how the build system works and packages are installed using pip. Now it will always create a wheel which contains installed data and during installation it is simply unpacked to the destination so we need to figure out prefix during runtime.
- I did not made the changes but according to documentation we must not use `setup.py` directly which means we would have to drop our custom commands and implement it some other way.

Because of the limitation and IMHO backwards step in the python build system I've also created PR https://github.com/virt-manager/virt-manager/pull/742 that introduces Meson instead of setuptools.

Fixes: https://github.com/virt-manager/virt-manager/issues/652

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,10 +11,5 @@
 
 /data/gschemas.compiled
 
-/man/virt-manager.1
-/man/virt-install.1
-/man/virt-clone.1
-/man/virt-xml.1
-
 /virt-manager.spec
 /virtinst/build.cfg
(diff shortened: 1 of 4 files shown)


phrdina:
Closing this PR as we agreed we will switch to meson.

PR #742, opened 2024-10-25
Introduce meson as our build system

phrdina:
As stated in the PR https://github.com/virt-manager/virt-manager/pull/741 the new python build system comes with some limitations, due to that I'm proposing to remove setuptools and adopt meson instead. I did check a lot of other python packages using GTK and most of them are using meson as their build system.

For the sake of review I've created multiple commits but I would probably prefer squashing them into single commit before pushing as both build systems will not work for most commits within this PR.


crobinso:
Nice work, this looks pretty thorough. I'm ok with switching to meson.

I don't fully understand pep 517, but, thinking outloud: virt-manager doesn't really fit the standard pip/setuptools mold because of dependence on a bunch of non-python system libs, build time configuration options, and needing to run external tools as part of install process like for glib schemas etc. We could make all that work with modern python infra, but it would be a bunch of build system disruption, so why not instead go all the way to meson which is more suited for a gtk application. Seems reasonable to me

Do you know any python UI apps that use meson? I see [meld](https://gitlab.gnome.org/GNOME/meld) does but it also has setup.py too so I don't quite follow that. The other app I tend to use as reference is [quodlibet](https://github.com/quodlibet/quodlibet) but that does not use meson.

I am a bit nervous in cramming this in right before a release. If there's something we miss that other distros hit we may have to do some quick fix follow up releases, but if you are handling release management then I'll leave the call up to you :)

Smaller comments:

* This preserves the simple path of git checkout + `./virt-install` which I appreciate. I would hate to lose that.
* Does meson let you use glob for scooping up files? Honestly having to list individual .py and .ui files is annoying IMO
* I think meson has native support for glib schemas and icon cache in the `gnome` module but that could come later

CC @pinotree @berrange for any general feedback on switching setuptools -> meson

phrdina:
> Nice work, this looks pretty thorough. I'm ok with switching to meson.
> 
> I don't fully understand pep 517, but, thinking outloud: virt-manager doesn't really fit the standard pip/setuptools mold because of dependence on a bunch of non-python system libs, build time configuration options, and needing to run external tools as part of install process like for glib schemas etc. We could make all that work with modern python infra, but it would be a bunch of build system disruption, so why not instead go all the way to meson which is more suited for a gtk application. Seems reasonable to me

Completely agree, we could make it work with pep 517 using setuptools but as you wrote virt-manager doesn't fit the standard pip ecosystem.

> Do you know any python UI apps that use meson? I see [meld](https://gitlab.gnome.org/GNOME/meld) does but it also has setup.py too so I don't quite follow that. The other app I tend to use as reference is [quodlibet](https://github.com/quodlibet/quodlibet) but that does not use meson.

I checked Fedora packages to see what the situation is and the following list uses meson for python UI apps:
- https://github.com/maoschanz/drawing
- https://github.com/maoschanz/dynamic-wallpaper-editor
- https://gitlab.gnome.org/World/gfeeds
- https://gitlab.gnome.org/GNOME/gnome-music
- https://gitlab.gnome.org/GNOME/gnome-tweaks
- https://gitlab.com/leinardi/gst

It's not a complete list but it seems to be pretty common to use meson for python GUI apps.

> 
> I am a bit nervous in cramming this in right before a release. If there's something we miss that other distros hit we may have to do some quick fix follow up releases, but if you are handling release management then I'll leave the call up to you :)

Good point, I'll think about it and most likely try bunch of distros to see if things are working correctly.

> Smaller comments:
> 
>     * This preserves the simple path of git checkout + `./virt-install` which I appreciate. I would hate to lose that.

Agree, it's nice feature and we should keep it.

>     * Does meson let you use glob for scooping up files? Honestly having to list individual .py and .ui files is annoying IMO

That's a design decision for meson, they don't support using glob, you need to always list all files explicitly https://mesonbuild.com/FAQ.html#why-cant-i-specify-target-files-with-a-wildcard.

>     * I think meson has native support for glib schemas and icon cache in the `gnome` module but that could come later

It does and commit ` meson: add icons and schemas options` is using the `gnome` module :).

Thanks for the review. I'll wait to give time to others and also do bit more testing and see if we will include it in the upcoming release or not.

eszlari:
> Do you know any python UI apps that use meson?

https://github.com/bottlesdevs/Bottles
https://github.com/lutris/lutris

berrange:
FYI, there is some contextually relevant discussion about switching to meson for libvirt-python too: https://gitlab.com/libvirt/libvirt-python/-/issues/21

My general view is that since virt-manager is not providing a public python API, and not being distributed or installed from  PyPI, there's no obvious benefit from using setuptools, and plenty of downsides. IOW, as an standalone application, I think it makes sense to leverage meson with virt-manager.

crobinso (approved):
OK I've approved this, please merge whenever you feel it's ready

PR #745, opened 2024-10-28
Update libvirt.org links

gszy:
Fix no longer valid URL fragments, wrap at 80th column, remove dots that follow only a few links.


ptoscano (approved):
Thanks!

Issue #747, opened 2024-11-05
switch console default to `Scale Display`->`Always`


crobinso:
Current default for spice/vnc `Scale Display` is 'Only when Fullscreen', so console scaling is not used when the VM is in windowed mode. If the VM display is smaller than the window, there's a black border around it. When VM display is larger, scrolledwindow bars pop up for navigating to the offscreen portion.

I think we should default to scaling instead, like virt-viewer does. It makes for a nicer out of the box experience IMO and simplifies a lot of cases. Examples:

* New VMs. qemu has defaulted to 1280x800 resolution for a while, but virt-manager's VM window was defaulting to 1024x768 resolution, so most newly created VMs would exceed the window size and show scrollbars, basically necessitating `Resize to VM`. I fixed the default VM window size, but that's not future proof, and does not apply for all VM configs.

* Existing VMs created outside of virt-manager. First time virt-manager sees a VM, it still defaults to the stock window size. If VM is has 1920 resolution for example, we would show scrollbars. If scaling is enabled, the VM will be scaled down and smooshed, but at least you can interact with it without requiring `Resize to VM` or touching scrollbars.

* When the host is configured with fractional scaling, it's kind of impossible for virt-manager to 
do anything correctly without enabling display scaling. There's more details here: https://github.com/virt-manager/virt-manager/issues/524

* When scaling is `Never` and the window is larger than the VM resolution (like if the host window is maximized while installing a new VM), it's difficult to interact with the edges of the VM screen.

Some of the above bits could be handled better for scaling `Never` case, like if we made the window smarter about following VM resolution changes, kind of how virt-viewer does it. But this would take some careful testing IMO, especially since virt-manager shares 'Console' and 'Details' in a single window. IMO it's simpler for now to just flip the scaling default.

The main downside of scaling is that the viewer will look blurry/not sharp when scaling is happening. For many usecases, users hate this. They want a VM window that perfectly fits the guest resolution, making everything look sharp. They can disable scaling, or adjust the window to the correct size with `Resize to VM`. Note, the current default of `Never` often doesn't work out of the box for them either, since a scrolledwindow will be shown, so they still need to manually invoke `Resize to VM` when resolution changes. So in that respect, the situation doesn't change much for them.

phrdina:
Fixed by #748 .

PR #748, opened 2024-11-05
switch console default to Scale Display->Always

crobinso:
I think this is a better default than the current `Only when Fullscreen`, which is the same as `Never` for VM windowed mode.

I wrote my reasoning here:
https://github.com/virt-manager/virt-manager/issues/747

### DIFF
--- a/data/org.virt-manager.virt-manager.gschema.xml
+++ b/data/org.virt-manager.virt-manager.gschema.xml
@@ -206,7 +206,7 @@
   <schema id="org.virt-manager.virt-manager.console"
           path="/org/virt-manager/virt-manager/console/">
     <key name="scaling" type="i">
-      <default>1</default>
+      <default>2</default>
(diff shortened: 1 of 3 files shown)


phrdina:
I personally don't like scaling but that is my preference. It probably makes sense to unify the behavior with virt-viewer and as you described in the issue it makes the user experience a bit better.

crobinso:
> I personally don't like scaling but that is my preference. It probably makes sense to unify the behavior with virt-viewer and as you described in the issue it makes the user experience a bit better.

@phrdina for scaling=never, do you use/like the scrolledwindow, or do you instead `Resize to VM` so you never see scrollbars?

phrdina:
Whenever possible I use `Resize to VM` and if that is not possible I have scaling=never and set the resolution to some reasonable value and usually maximize the VM window to not have any scrollbars.

PR #749, opened 2024-11-08
Typo in en_GB.po

Pompilos:
A typo (duplicate "be") in string

### DIFF
--- a/po/en_GB.po
+++ b/po/en_GB.po
@@ -5950,8 +5950,8 @@ msgid "location kernel/initrd may only be specified with a location URL/path"
 msgstr "location kernel/initrd may only be specified with a location URL/path"
 
 #: virtinst/install/installertreemedia.py:119
-msgid "location kernel/initrd must be be specified as a pair"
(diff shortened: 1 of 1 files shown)


sunwire:
Good catch but it have to be corrected first in [this line](https://github.com/virt-manager/virt-manager/blob/c99b64725ca7cfd90e666b4fee87659531ba3d39/virtinst/install/installertreemedia.py#L119)

Pompilos:
> Good catch but it have to be corrected first in [this line](https://github.com/virt-manager/virt-manager/blob/c99b64725ca7cfd90e666b4fee87659531ba3d39/virtinst/install/installertreemedia.py#L119)

Thank you for pointing it out. I have proposed to fix it as well.

ptoscano:
Hi @Pompilos,

translations are handled using Weblate, and they are pulled from it rather than being modified directly in the sources. See the `Translations` section in `CONTRIBUTING.md`. Hence, please fix the the translation in Weblate.

Thanks anyway for the PR.

PR #750, opened 2024-11-10
Update installertreemedia.py

Pompilos:
Removing duplicated verb "be"


phrdina:
Thanks for the patch. I've fixed the test case that checks for the updated string.

Issue #692, opened 2024-07-18
RFE: need virt-install `--boot bios` to override UEFI default


andreabolognani:
With more and more operating systems becoming UEFI-first (if not outright UEFI-only), we can expect that at some point in the not-so-distant future virt-install will start preferring it to BIOS. When that happens, we will need a way to opt out of UEFI.

The obvious expectation on the user's part would probably be to be able to use `--boot bios`, mirroring `--boot uefi` that can currently be used to opt into the non-default firmware choice. That option doesn't currently exist.

What *does* exist is `--boot firmware=bios`. That's a bit more verbose and less discoverable, but that in itself would probably not be a big issue. The problem is that it doesn't really work right now:

```
$ virt-install --name cirros-bios --osinfo linux2022 --boot firmware=bios --import --disk /var/lib/libvirt/images/cirros.qcow2 
Starting install...
ERROR    operation failed: Unable to find 'bios' firmware that is compatible with the current configuration
```

It's easy to see why this happens: `--boot firmware=bios` translates to `<os firmware='bios'>`; however, despite having the SeaBIOS package installed on my system, there is no firmware descriptor for any of the files it contains, so libvirt is unable to come up with a reasonable match for the requirements.

We could push for distro packages to introduce such a firmware descriptor, making this argument somewhat moot, but I also think that we might want to have a way to just explicitly generate an empty `<os>` element, as a way to really maintain the current/historical behavior.

While poking around a bit, I've also noticed that `--boot uefi` and `--boot firmware=uefi` seem to have slightly different behaviors. I'm not yet sure whether the two can be reconciled, but if at all possible I think we should aim for that as the current situation can be a significant source of confusion IMO.

Before I start sinking time into this, I'd like to hear the maintainers' opinions on the topic, specifically:

* would introducing `--boot bios` be welcome?
* should `--boot uefi` and `--boot firmware=uefi` behave the same way?
* should there be a way to generate `<os>` with no firmware information, and if so what would be a good CLI for it?

Thanks!

phrdina:
`--boot uefi` and `--boot firmware=efi` are not equivalent options. The former is a shortcut to request default UEFI firmware for a VM which will make necessary changes to the VM XML. The letter directly translates to VM XML attribute and does nothing else. Nowadays it behaves the same for libvirt that supports firmware auto-selection but before that features was introduced in libvirt it was getting loader path from libvirt capabilities and setting necessary XML bits.

[...]

PR #759, opened 2024-11-14
Add `--boot uefi=off`

crobinso:
Add `--boot uefi=off` for opting out of uefi when virt-install would default to it, and for virt-xml to clear out all UEFI config for an existing VM

Resolves: https://github.com/virt-manager/virt-manager/issues/692

### DIFF
--- a/virtinst/guest.py
+++ b/virtinst/guest.py
@@ -221,7 +221,7 @@ class Guest(XMLBuilder):
         self.num_pcie_root_ports = 14
 
         self.skip_default_osinfo = False
-        self.uefi_requested = False
+        self.uefi_requested = None
         self.__osinfo = None
         self._capsinfo = None
         self._domcaps = None
@@ -1038,6 +1038,9 @@ class Guest(XMLBuilder):
         return path
 
     def _set_default_uefi(self):
+        if self.uefi_requested is False:
+            return
+
         use_default_uefi = (self.prefers_uefi() and
             not self.os.kernel and
             not self.os.loader and
--- a/virtinst/guest.py
+++ b/virtinst/guest.py
@@ -600,6 +600,22 @@ class Guest(XMLBuilder):
         log.debug("Setting default UEFI path=%s", path)
         self.set_uefi_path(path)
 
+    def disable_uefi(self):
+        self.os.firmware = None
+        self.os.loader = None
+        self.os.loader_ro = None
+        self.os.loader_type = None
+        self.os.loader_secure = None
+        self.os.nvram = None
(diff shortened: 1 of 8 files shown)


phrdina (changes requested):
Overall looks good and makes sense to implement this, but the `virt-xml` bit needs to be fixed.

virtinst/guest.py
```diff
@@ -600,6 +600,17 @@ def enable_uefi(self):
...
         self.set_uefi_path(path)
 
+    def disable_uefi(self):
+        self.os.firmware = None
+        self.os.loader = None
+        self.os.loader_ro = None
+        self.os.loader_type = None
+        self.os.loader_secure = None
+        self.os.nvram = None
+        self.os.nvram_template = None
```
phrdina, line 610: Unfortunately this still doesn't work. Virt-install currently doesn't know about `format` attributes so it doesn't remove them and libvirt fails with this error:
```
ERROR    no loader path specified and firmware auto selection disabled
```

The --print-diff output looks like this with the current code:

```xml
--- Original XML
+++ Altered XML
@@ -9,14 +9,11 @@
   <memory unit="KiB">8388608</memory>
   <currentMemory unit="KiB">8388608</currentMemory>
   <vcpu placement="static">8</vcpu>
-  <os firmware="efi">
+  <os>
     <type arch="x86_64" machine="pc-q35-9.1">hvm</type>
-    <firmware>
-      <feature enabled="yes" name="enrolled-keys"/>
-      <feature enabled="yes" name="secure-boot"/>
-    </firmware>
-    <loader readonly="yes" secure="yes" type="pflash" format="qcow2">/usr/share/edk2/ovmf/OVMF_CODE_4M.secboot.qcow2</loader>
-    <nvram template="/usr/share/edk2/ovmf/OVMF_VARS_4M.secboot.qcow2" format="qcow2">/var/lib/libvirt/qemu/nvram/fedora40_VARS.qcow2</nvram>
+    <firmware/>
+    <loader format="qcow2"/>
+    <nvram format="qcow2"/>
     <boot dev="hd"/>
   </os>
   <features>
```

We need to add `format` into both `loader` and `nvram` or rework both to `XMLChildProperty`. Not sure if there is any other easy fix.

crobinso: aww fooey, was testing with an old VM without format='qcow2'. but this makes me realize it's hard to be future proof here. there's other bits we aren't covering like ./nvram/@type and ./loader/@stateless. Latest patch will clear all properties we are tracking, and then force removes the nvram, loader, and firmware XML nodes, so hopefully that covers us.

Issue #751, opened 2024-11-11
RFE: Convenience option for passt port forwarding


andreabolognani:
With #697 merged, it is now possible to create VMs that use passt for networking.

Unfortunately, due to how the libvirt XML is designed and how virt-install maps its structure to command line options, doing something as common as making a guest port accessible to the host requires fairly unwieldy invocations, for example:

```
--network backend.type=passt,portForward0.proto=tcp,portForward0.range0.start=8080,portForward0.range0.to=80
```

For comparison's sake, when using podman the same result can be achieved with

```
-p 8080:80
```

It would be nice to introduce a shorthand syntax that makes at least the straightforward scenarios more accessible. For example it could look like

```
--network passt,port-forward=8080:80
```

andreabolognani:
@phrdina since you were the one who implemented passt support in the first place, do you have any thoughts about this idea? Or possibly, maybe, by any chance, even a shred of interest in working on it yourself? O:-)

crobinso:
I submitted this: https://github.com/virt-manager/virt-manager/pull/765  . Example: `--network backend.type=passt,portForward=123:456,portForward1=127.0.0.1:8080:80/udp`

andreabolognani:
@crobinso thanks for looking into this!

I suppose the use of `portForward=...,portForward1=...` is needed to tell the various instances apart. Is this a technical question, i.e. virt-install's CLI parser is completely unable to tell two instances of the same parameter apart, or a matter of consistency with how other options are handled?

As you can probably tell I'm not very familiar with this part of the code, but I see that a callback is involved, which naively makes me hope that perhaps it would be possible to make things even more user-friendly by removing the need to number instances?

Similarly, in the interest of making things as convenient for users as possible, is there any chance that we could support `--network passt` as a shorter alias for `--network backend.type=passt`, or is that out of the question?

Don't get me wrong, what you've implemented is already leaps and bounds more usable than the 1:1 mapping to XML elements that was there before! But if we could somehow manage to accept a more svelte `--network passt,portForward=123:456,portForward=127.0.0.1:8080:80/udp` I think we would *really* be on to a winner!

andreabolognani:
P.S. if we start accepting bespoke syntax that's not 1:1 mapped to XML elements we might as well call this either `forward`, `ports` (following passt's lead) or `publish` (following that of docker/podman).

[...]

PR #765, opened 2024-11-16
cli: add `--network portForward=8080:80` convenience syntax

crobinso:
Roughly mirroring `podman run -p` syntax. Example:

--network backend.type=passt,portForward=123:456,portForward1=127.0.0.1:8080:80/udp

Resolves: https://github.com/virt-manager/virt-manager/issues/751

### DIFF
--- a/man/virt-install.rst
+++ b/man/virt-install.rst
@@ -1338,6 +1338,20 @@ Some example suboptions:
 
     For ``HOSTDEV`` format, see ``--hostdev`` documentation
 
+``portForward=[ADDRESS:]HOSTPORT[:GUESTPORT][/PROTO]``
+    Simpler option for specifying port forwarding with
+    ``--network passt`` networks. Roughly matches ``podman run -p``
+    syntax. HOSTPORT can be a represented as a range like ``7000-8000``, but
+    GUESTPORT can only be a single port. If GUESTPORT is not provided, host
+    and guest ports are assumed to match.
+
+    Examples:
+
+    .. code-block::
+
+           --network passt,portForward=8080:80 \
+           --network passt,portForward0=7000-8000/udp,portForward1=127.0.0.1:2222:22 \
+
 
 GRAPHICS OPTIONS
 ================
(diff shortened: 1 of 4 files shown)


man/virt-install.rst
```diff
@@ -1338,6 +1338,15 @@ Some example suboptions:
 
     For ``HOSTDEV`` format, see ``--hostdev`` documentation
 
+``portForward=[ADDRESS:]GUESTPORT:HOSTPORT[/PROTO]``
```
andreabolognani, line 1341: I believe this should be `[ADDRESS:][HOSTPORT:]GUESTPORT[/PROTO]`.

The host part definitely comes first, otherwise `8080:80` wouldn't make much sense: you'd be mapping an unprivileged guest port to a privileged host one, defeating the purpose of using unprivileged guest networking in the first place.

And the `HOSTPORT` should be marked as optional, since libvirt only requires the host part (`start`) to be provided unless the guest part (`to`) doesn't match. This would be *almost* consistent with how the corresponding docker/podman feature works.

I say "almost" because podman/docker would react to `-p 80` by mapping guest port 80 to a random unprivileged host port, whereas libvirt would react to `portForward=80` by attempting to map guest port 80 to... Host port 80, which is probably not what you wanted and will likely fail.

That is not virt-manager's problem though.

crobinso: Sounds good, will fix!

tests/test_cli.py
```diff
@@ -653,6 +653,8 @@ def add_compare(self, cat, args, compbase, **kwargs):
 --network user,address.type=ccw,address.cssid=0xfe,address.ssid=0,address.devno=01,boot.order=15,boot.loadparm=SYSTEM1
 --network model=vmxnet3
 --network backend.type=passt,backend.logFile=/tmp/foo.log,portForward0.proto=tcp,portForward0.address=192.168.10.10,portForward0.dev=eth0,portForward0.range0.start=4000,portForward0.range0.end=5000,portForward0.range0.to=10000,portForward0.range0.exclude=no,portForward0.range1.start=6000,portForward1.proto=tcp,portForward1.range0.start=2022,portForward1.range0.to=22
+--network passt,portForward=8080:80
+--network passt,portForward0=7000-8000/udp,portForward1=127.0.0.1:2222:22
```
andreabolognani, line 657: This could probably use a bit more coverage, especially for the less common combinations. For example:

```
--network passt,portForward0=2001:db8:ac10:fd01::1:10:3000-4000:30,portForward1=127.0.0.1:5000-6000:50
```

We're also missing coverage for the very simplest case, such as:

```
--network passt,portForward=8080
```

It all seems to work fine right now! I'm just concerned that it might accidentally get broken in the future.

andreabolognani (commented):
@crobinso thanks! Unfortunately I have a couple more comments O:-)

tests/test_cli.py
crobinso: ACK, I think it looks good now

PR #779, opened 2024-12-03
addhardware:panic: Fill in model combo with advertised values by libvirt

morecache:


ui/addhardware.ui
```diff
@@ -1318,12 +1318,6 @@
                                   <object class="GtkComboBox" id="panic-model">
                                     <property name="visible">True</property>
                                     <property name="can-focus">False</property>
-                                    <property name="has-entry">True</property>
-                                    <child internal-child="entry">
-                                      <object class="GtkEntry">
-                                        <property name="can-focus">True</property>
-                                      </object>
-                                    </child>
```
phrdina, line 1326: I would drop this change. We should still allow users to type in the model manually in case they connect to libvirt that doesn't report panic models.

virtManager/addhardware.py
```diff
@@ -860,7 +860,11 @@ def _build_redir_type_combo(self):
 
 
     def _build_panic_model_combo(self):
-        values = [[None, _("Hypervisor default")]]
+        guest = self.vm.get_xmlobj()
+        values = []
+        for m in guest.lookup_domcaps().supported_panic_models():
+            values.append([m, m])
+
```
phrdina, line 867: For the same reason we should keep the `Hypervisor default` and set it as default_value.

phrdina: I've added a line that will set `Hypervisor default` as the default value and pushed it. Once tests are done I'll merge this. Thanks!

PR #794, opened 2024-12-22
Better error message for inactive networks

Two-Play:
It's very confusing for beginners when they start a VM and an error occurs that it won't work because the network is inactive. That's why I've added a solution to this problem in the error message.

The best variant would probably be to ask if the network should be activated, but unfortunately, I don't have enough knowledge and time for that yet.

### DIFF
--- a/virtManager/asyncjob.py
+++ b/virtManager/asyncjob.py
@@ -76,6 +76,16 @@ def cb_wrapper(callback, asyncjob, *args, **kwargs):
             asyncjob.job_canceled):
             return  # pragma: no cover
 
+        # If network is inactive, show a different error message
+        if (isinstance(e, libvirt.libvirtError) and
+                e.get_error_code() == libvirt.VIR_ERR_OPERATION_INVALID and
+                e.get_error_domain() == libvirt.VIR_FROM_NETWORK):
+            asyncjob.set_error(
+                str(e) + "\n\n" +
+                _("Go to: Edit -> Connection Details -> Network Interfaces to activate the network."),
+                "".join(traceback.format_exc()))
+            return # pragma: no cover
+
         asyncjob.set_error(str(e), "".join(traceback.format_exc()))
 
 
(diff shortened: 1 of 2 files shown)


virtManager/manager.py
```diff
@@ -171,6 +171,13 @@ def __init__(self):
         for conn in connmanager.conns.values():
             self._conn_added(connmanager, conn)
 
+        # Select the first item in the list if it exists
+        vmlist = self.widget("vm-list")
+        model = vmlist.get_model()
+        if len(model) > 0:
+            selection = vmlist.get_selection()
+            selection.select_path(Gtk.TreePath.new_first())
+
```
phrdina, line 180: This change is unrelated to the PR so it should not be part of it at all and I don't see any justification to make such change.

virtManager/asyncjob.py
```diff
@@ -76,6 +76,16 @@ def cb_wrapper(callback, asyncjob, *args, **kwargs):
...
+        # If network is inactive, show a different error message
+        if (isinstance(e, libvirt.libvirtError) and
+                e.get_error_code() == libvirt.VIR_ERR_OPERATION_INVALID and
+                e.get_error_domain() == libvirt.VIR_FROM_NETWORK):
+            asyncjob.set_error(
+                str(e) + "\n\n" +
+                _("Go to: Edit -> Connection Details -> Network Interfaces to activate the network."),
+                "".join(traceback.format_exc()))
+            return # pragma: no cover
+
```
phrdina, line 88: This works but it's not correct way to do it. getting error_domain `VIR_FROM_NETWORK` and error_code `VIR_ERR_OPERATION_INVALID` doesn't guarantee that it is due to not-activated network.

phrdina (commented):
I'll close this PR as we will not merge it in this state.

We could possibly add a code that would check the VM definition if any libvirt network is configured for that VM and check relevant networks if they are active and warn the user, but to be completely honest not sure if anybody else will have time to implement this improvement.

PR #804, opened 2024-12-30
virt-install: Add some new features

morecache:


phrdina:
I've changed `templateformat` to `templateFormat` to match libvirt XML.

PR #808, opened 2025-01-05
Add couple of features to virt-install

morecache:


phrdina:
I've changed `dynamicmemslots` to `dynamicMemslots` to match attribute name from libvirt XML.

PR #809, opened 2025-01-05
feat(virt-install): add ability to grab cloud-init user-data from url

syaghoubi00:
Adds feature to grab `cloud-init` `user-data` from a remote source, so the user doesn't have to have a local copy of the file.


ptoscano:
Hmm it looks like most of the additions here are reimplementing what is already available in `virtinst/install/urlfetcher.py`; would you please take a look at that?

syaghoubi00:
Went ahead and used the existing code. Not sure how I missed that, thanks for pointing it out.

The `base_url` feels a little hacky, not sure what the best way to handle that is.

I believe `user-data` from `ftp://` should work now as well, since that is supported in `fetcherForURI`, so that's a nice bonus.

phrdina:
Hi, thanks for your contribution.

This adds URL support only to the `user-data` option, we should add the same to `meta-data` and `network-config`.

I've created a new PR refactoring the current code and adding URL support for all options https://github.com/virt-manager/virt-manager/pull/873 .

This PR has few issues. The main one is that it still contains the original commits, revert of the original commit, new commit and merge commit. In most cases you want to remove unnecessary commits and make changes to existing commits if the project doesn't use "squash merging" of PR.

PR #818, opened 2025-01-12
snapshots: Default to internal snapshots in some cases

andreabolognani:
Avoid creating snapshots that are (currently) impossible to delete.


phrdina:
Hi, thanks for the patch. I was planning to implement it slightly differently, base the internal vs external snapshot type on the currently active snapshot. I'm also planning to add a warning in case user decides to mix snapshot types.

andreabolognani:
@phrdina this is the quick and dirty version that I hacked together on a Sunday morning after hitting the problem myself, mostly to make sure that we wouldn't lose track of it. If you are planning to work on a more polished implementation, by all means do so! You have my full support :)

PR #824, opened 2025-01-21
Pick better default snapshot mode and warn users to not mix snapshot modes

phrdina:


virtManager/details/snapshots.py
```diff
@@ -352,6 +370,16 @@ def _name_changed_cb(self, src):
         self._populate_memory_path()
 
     def _ok_clicked_cb(self, src):
+        current_mode = self._get_current_mode()
+
+        if current_mode and current_mode != self._get_mode():
+            result = self.err.yes_no(_("Using external and internal snapshot mode "
```
andreabolognani, line 376: Perhaps "Mixing external and internal snaphots for the same VM..."

virtManager/details/snapshots.py
```diff
@@ -342,6 +349,17 @@ def _create_new_snapshot(self):
...
+    def _get_current_mode(self):
+        current = self.vm.get_current_snapshot()
+
+        if current is None:
+            return None
+
+        if current.is_external():
+            return "external"
+
+        return "internal"
```
andreabolognani, line 361: The fact that the mode is returned as a verbatim string as opposed to something like an enum or even just a string constant is a bit icky. I know it's pre-existing, but perhaps you could consider changing that, possibly in a follow-up PR.

andreabolognani (commented):
@phrdina added a couple nits but overall looks good and seems to work fine. Thanks a lot for looking into it!

virtManager/details/snapshots.py
phrdina: Agreed that it's icky, I'll leave it to follow-up PR.

PR #825, opened 2025-01-21
createvm: prioritize riscv64

xypron:
As all major distros support the riscv64 architecture, add it to the list of prioritized architectures. This will move it up in the architecture drop-down menu and thereby easier to find when creating a new VM.


crobinso:
Good call, thanks!

PR #827, opened 2025-01-22
feat: add replace space to _ in snapshot

0xlildoudou:
Hi,

because it's a real pain to get an error when you put spaces in names


phrdina:
Hi, can you please provide some example where the space causes issue for you?

0xlildoudou:
For example, VMware uses IDs for snapshots, so you can put spaces in the name.

With virt-manager you can't (as in the Linux system as a whole). However, making a simple replace to change the space to a _ avoids a virt-manager error (which sometimes means having to redo the description, etc.).

Personally, I've done it for snapshots and VM names.

Issue #722, opened 2024-09-12
RFE: refactor the code base with black


ptoscano:
Python relies a lot on indentation, and PEP 8 [1] already provides guidelines of a basic consistent code style; that said, there are many ways to still format a Python code base.

Since the last few years, a new PSF-backed project has been providing a consistent way to format Python code, using a tool called black [2]. black is a very opinionated formatter, and it does not allow for many tweaks (mostly line length, whether to normalize strings). In addition to that, black also offers a way to check whether the formatting is correct, which makes it a good fit also in CI jobs.

virt-manager does not have a consistent style; this is seen from time to time in contributions, as there is confusion on how to format bits. Hence this RFE: reformat the code base using black. The only needed steps would be to decide on a line length (the default of black is 88), and whether normalize strings (the default is yes); in case any of the two would be needed, a simple `pyproject.toml` can be added with the config.

Formatting the code with black has a various benefits:

1. a unified code style across virt-manager; no more differences even within the same file
1. a simple way to format the code base: run `black` on the whole source directory, on specific files, etc, and you are good to go
1. no more need to think about the formatting when doing changes, simply run `black` before committing (it can be also automated as local pre-commit hook)
   - this also simplifies the contributions in PRs
1. a way to check that the style is correct: `black --check` does everything needed
   - this can be integrated as CI job
   - the `psf/black@stable` github action greatly helps in this
1. the unified style will basically get rid of all the `flake8` issues related to formatting
   - because of the `black` formatting, E203 and W503 may be reported by `flake8`, however it is easy to ignore them

There are also few downsides:
1. there would be a giant commit with lots of formatting changes
   - the "pollution" of `git blame` can be avoided by using a file with revisions to ignore; see also https://gist.github.com/kateinoigakukun/b0bc920e587851bfffa98b9e279175f2 for an example
1. backports (usually done by distros) to versions released before the reformatting would be harder, as most likely they will require manual rewrites
   - this would be less of a problem over time
1. the style sometimes it is not pretty to see, for example the famous "sad face":
   ```python
   if (
       # long condition
   ):
       # some code
   ```
   while not nice to see (especially if you are not used to it), in the end it's one of the results of the consistent placement of opening & closing brackets that `black` manages; also, from personal experience, the style tends to be formatted "ugly" when it is already complex enough, for example lots of nested function calls, and usually becomes nicer when simplified/split a bit

[...]

PR #834, opened 2025-01-30
Reformat the codebase with black

ptoscano:
Make use of the `black(1)` tool [1] to reformat the codebase. The only behaviour change for black is the different line length, i.e. 100 instead of black's 88.

There reformatting was split in two commits to ease the review of the changes and the navigation through git history. A followup of this PR will include the addition of a file with the revisions to ignore for `git blame`, and a github workflow to check the code style in PRs.

Closes: #722

[1] https://github.com/psf/black


phrdina:
Overall looks good, but there are some fixes needed.

pycodestyle complains in multiple files:
```
[W503] line break before binary operator
[E203] whitespace before ':'
```
we need to add these to ignore in `setup.cfg`.

pylint complains in a lot of files with:
```
W1404: Implicit string concatenation found in call (implicit-str-concat)
```
that will need to be fixed manually as it seems that black is not able to reformat it correctly.

Before merging I would probably make a single commit, that is recommended together with the `.git-blame-ignore-revs` file.

crobinso:
I think our use of pycodestyle should just be ripped out, it's largely redundant if we are using `black`.

@pinotree should this come with some CI magic to make sure style isn't regressing?

PR #839, opened 2025-02-11
Add pstore device support for storing oops/panic logs before the guest crashes

morecache:


crobinso:
Thanks! I adjusted code after the `black` reformat and pushed manually

PR #843, opened 2025-02-13
xmlapi: Don't escape property values

### LINKED ISSUE #838: virt-xml double encodes VNC passwords
Setting `"&"` as the VNC password leads to an actual password of `"&amp;"`.

```
# virt-xml --edit --graphics 'vnc,passwd=&' foo
# virsh dumpxml --security-info --inactive foo | grep graphics
    <graphics type='vnc' port='-1' autoport='yes' passwd='&amp;amp;'>
    </graphics>
```

Note that the `"&"` has turned into `"&amp;amp;"` in XML, while the correct value would be just `"&amp;"`. Some double entity encoding has taken place on the way to XML.

This can be also be observed with virt-manager: Shutdown machine, open "VNC Server" page, check "password" and "show password" boxes, type `&` into text field, hit "Apply".  The`&` turns into `&amp;` by itself.

Interestingly, the length limit is enforced on the encoded version:

```
# virt-xml --edit --graphics 'vnc,passwd=1234&' foo
ERROR    unsupported configuration: VNC password is 9 characters long, only 8 permitted
```

`"1234&"` is only 5 characters, but `"1234&amp;"` is 9.

So the misplaced entity encoding seems to be in front of length check, not after. Hopefully that helps to find it.

Thanks!

mvollmer:
libxml2 does that already.

Fixes #838


phrdina:
Nice catch. I did some testing and investigation what is the root cause and reading libxml2  @API we indeed shouldn't  escape the string when using `setProp` method, the other case where we escape the string is for `setContent` where the API explicitly states that here we need to escape. So this fix is correct.

For the `listen` issue that is a bug in libvirt as libvirt returns incorrect XML with the `&` not escaped correctly and virt-xml fails to parse the returned XML.

I've fixed the merge conflict introduced by reformatting code with black and once all the CI passes I'll merge this.

PR #856, opened 2025-03-02
tests: Increase virtio-mem block size

akihikodaki:
virtio-mem block size must be equal to or greater than the transparent huge page size; otherwise, libvirt raises an error and a test will fail. For example, on Asahi Linux, the transparent huge page size is 32 MiB, which is greater than 2 MiB, the specified virtio-mem block size.

On Linux 6.13, the configuration with the maximum transparent huge page size is Arm64 with 64 KiB, and it has 512 MiB transparent huge pages. Increase the block size to 512 MiB so that the test passes on every Linux configuration.


crobinso:
Thanks! I resolved the conflict locally and pushed, so closing this

PR #860, opened 2025-03-07
snapshots: Add order combobox in snapshot page

### LINKED ISSUE #847: thanks but make virt-manager and virsh output snapshots in chronological creation date order
thanks for virt-manager and virsh

using those tools daily

question: would it be a big problem to sort the list of snapshots by Creation Time?

in virsh this is still kind of manageable but in virt-manager it's not

 virsh snapshot-list --domain vmname
 Name                           Creation Time               State
---------------------------------------------------------------------
 snapshot2                      2025-02-11 20:23:40 +0100   shutoff
 snapshot4                      2025-02-14 10:35:44 +0100   shutoff
 snapshot5                      2025-02-19 10:43:12 +0100   shutoff
 snapshot3                      2025-02-14 10:25:01 +0100   shutoff
 snapshot1                      2025-02-06 19:23:20 +0100   shutoff

thanks

morecache:
The "model.set_sort_column_id(4, Gtk.SortType.ASCENDING)" in snapshots.py makes ascend order by column 4(snapshot name column). This patch added a combobox for soring, The default one is descending order by creation time of snapshots, So the latest one is always on the top.

E.g:
root@localhost:~ # virsh snapshot-list --domain tumbleweed
 Name            Creation Time               State
------------------------------------------------------
 desktop_ready   2025-03-03 14:26:23 +0000   running
 paused1         2025-03-03 14:56:57 +0000   paused
 poweroff1       2025-03-03 14:45:17 +0000   shutoff
 snapshot1       2025-03-03 14:23:17 +0000   running
 youtube         2025-03-03 14:42:19 +0000   running

We can see the snapshot list in snapshot page via UI:
 desktop_ready
 paused1
 poweroff1
 snapshot1
 youtube

After the patch, The default order option in the combobox is by creation time, So we can see the snapshot list in snapshot page via UI:
 paused1
 poweroff1
 youtube
 desktop_ready
 snapshot1

Fixes #847

It's sure thing that tree view/tree structure is the best solution, But it take times. 
Here is a simple workaround: A sorting combobox as a short-term solution. Once we got a fully functional tree structure for snapshot UI, We can remove/improve this combobox.

### DIFF
--- a/ui/snapshots.ui
+++ b/ui/snapshots.ui
@@ -365,6 +365,56 @@
           </packing>
         </child>
         <child>
+          <object class="GtkBox" id="snapshot-order-box">
+            <property name="visible">True</property>
+            <property name="can-focus">False</property>
+            <property name="spacing">6</property>
+            <child>
+              <object class="GtkBox" id="hbox11">
+                <property name="visible">True</property>
+                <property name="can-focus">False</property>
+                <child>
+                  <object class="GtkLabel" id="snapshot-order">
+                    <property name="visible">True</property>
+                    <property name="can-focus">False</property>
+                    <property name="halign">start</property>
+                    <property name="label" translatable="yes">Order by:</property>
+                    <property name="use-underline">True</property>
+                    <property name="mnemonic-widget">snapshot-orderby</property>
+                  </object>
+                  <packing>
+                    <property name="expand">False</property>
+                    <property name="fill">False</property>
+                    <property name="position">0</property>
+                  </packing>
+                </child>
+                <child>
+                  <object class="GtkComboBox" id="snapshot-orderby">
+                    <property name="visible">True</property>
+                    <property name="can-focus">False</property>
+                    <signal name="changed" handler="on_snapshot_orderby_changed" swapped="no"/>
+                  </object>
+                  <packing>
+                    <property name="expand">False</property>
+                    <property name="fill">False</property>
+                    <property name="position">1</property>
+                  </packing>
+                </child>
+              </object>
+              <packing>
+                <property name="expand">True</property>
+                <property name="fill">True</property>
+                <property name="position">0</property>
+              </packing>
+            </child>
+          </object>
+          <packing>
+            <property name="expand">False</property>
+            <property name="fill">False</property>
+            <property name="position">1</property>
+          </packing>
+        </child>
+        <child>
           <object class="GtkBox" id="box2">
             <property name="visible">True</property>
             <property name="can-focus">False</property>
@@ -528,7 +578,7 @@
           <packing>
             <property name="expand">False</property>
             <property name="fill">True</property>
-            <property name="position">1</property>
+            <property name="position">2</property>
           </packing>
         </child>
       </object>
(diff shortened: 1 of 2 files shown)


phrdina:
Thanks for working on this but it doesn't feel like a complete solution.

There are only two option and they don't explain how the ordering is done. For time it uses descending and for name it uses ascending. If we want to have new UI elements to control ordering we should allow users to pick the direction and the field separately.

We had snapshot without ordering for the whole time, I don't see any reason to quickly add some workaround.

morecache:
All right.

This 'solution' is indeed rather basic.
However, as the amount of user snapshots increases, the current design of sorting snapshots in alphabetical order will make it increasingly difficult for users to quickly locate their snapshots. 
For instance, the most recent snapshot might not be displayed at the top, but instead be buried among several older snapshots.

* If we want to alleviate this issue, I believe I can improve this patch.
* If we want to completely resolve this issue, I think we should use a tree structure, but this will likely require a complete overhaul of the original layout of the snapshot page. If we simply add a tree structure on top of the existing page layout without making such changes, the user experience will deteriorate significantly as the amount of snapshot branches and descendants increases.

May I have your idea? @phrdina

phrdina:
I would go for the tree view as an example something like this:

Screenshot at 2025-03-07 15-49-31

It should probably have colapse/expand all buttons but the default state would be expand all.

I don't use snapshots so if there is something else we should add there let's discuss it here and figure out the design before any work is started on it.

phrdina:
> The example looks nice!
> 
> Don't know whether the example already supports below items or if they are already planned, Just a reminder:
> 
>     * support multi-select deletion.

That works as it uses the same gtk selection as it used before, only different representation of the same data.

>     * Using the checkmark icon to identify the latest snapshot is not enough, we should add a fake, unclickable node to highlight the current location.

I'll think about it how to handle this as it can be useful to have it there but technically the current state is not a snapshot and in case there are no snapshots not sure if we should list this node or leave it empty.

> 
>     * After deletion, All affected subordinate nodes(including the current location node) must be correctly linked to the proper parent node.

It always gets the actual data from libvirt so any deletion using virt-manager triggers refresh of the view.

>     * If we use the vertical tree view, the current Snapshot-list/Details footprint is fine. If we switch to a horizontal tree view(which is my personal prefer), the current footprint will need to be modified because it makes more sense to reserve most of the screen space for the snapshot list area, which allows users to visualize the snapshot chain more intuitively, the Details area only needs a small amount of screen space in this case.

I don't think that using horizontal tree view is possible with GTK. There is GtkTreeView class and I don't see any API that would allow setting vertical vs horizontal. Anyway agreed that we can make the details part smaller.

>     * The line "VM State:" in the Snapshot-list area doesn't make much sense and could be removed, we could just use the appropriate icon to identify the VM state of this snapshot. As for external/internal, I think it could be added after the snapshot name and surround it with parentheses.

Agreed that we can remove the VM state from tree view, the icon already changes. I would still keep the external/internal on separate line using smaller font to make it separate from the snapshot name.

rebroad:
> I would go for the tree view as an example something like this:
> 
> Screenshot at 2025-03-07 15-49-31
> 
> It should probably have colapse/expand all buttons but the default state would be expand all.
> 
> I don't use snapshots so if there is something else we should add there let's discuss it here and figure out the design before any work is started on it.

Does the code exist for this tree-view?

PR #868, opened 2025-03-12
rename `default-hvs` to `hvs-choices`

fkobi:
IMO "default" implies that the choice be changed and can from what I gather it is not the case.

For the codebase to be coherent I have also changed not only the build system but also the source code.


phrdina:
What is your motivation to change the naming? I personally don't see value renaming the option as the description in `meson_options.txt` already describes what it does and this changes adds extra work to downstream packagers.

You could also say it's a default list as you can still provide custom URI when creating new connection.

phrdina:
I agree that the name could be better but it is like that since it was introduced and I don't think renaming it and basically adding extra work to every package maintainer for every distribution using virt-manager is worth it.

phrdina:
Closing this PR as I don't think the value renaming the build options justifies the extra work introduced to every maintainer of every distribution already using this option.

PR #873, opened 2025-03-17
Add support to get cloud-init config file from URL

phrdina:
This allows users to specify `meta-data`, `user-data` and `network-config` using URL in addition to local file.

### DIFF
--- a/man/virt-install.rst
+++ b/man/virt-install.rst
@@ -897,10 +897,12 @@ Sub options are:
 ``meta-data=``
     Specify a cloud-init meta-data file to add directly to the iso. All other
     meta-data configuration options on the --cloud-init command line are ignored.
+    Can be path to local file or URL.
 
 ``user-data=``
     Specify a cloud-init user-data file to add directly to the iso. All other
     user-data configuration options on the --cloud-init command line are ignored.
+    Can be path to local file or URL.
 
 ``root-ssh-key=``
     Specify a public key to inject into the guest, providing ssh access to the
@@ -914,6 +916,7 @@ Sub options are:
 
 ``network-config=``
     Specify a cloud-init network-config file to add directly to the iso.
+    Can be path to local file or URL.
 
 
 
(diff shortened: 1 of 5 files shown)


man/virt-install.rst
```diff
@@ -897,10 +897,12 @@ Sub options are:
 ``meta-data=``
     Specify a cloud-init meta-data file to add directly to the iso. All other
     meta-data configuration options on the --cloud-init command line are ignored.
+    Can by path to local file or URL.
```
crobinso, line 900: `Can by` -> `Can be`, same with other files

phrdina: Will fix that, thanks.

phrdina:
> Looks good to me but it's a red flag there's no test/ additions, even if coverage is clean. We want to make sure future refactors don't cause at least accepting URL to regress.
> 
> I think this should cover it (tested but need to commit new XML file and probably run `black`)
> 
> ```diff
> diff --git a/tests/test_cli.py b/tests/test_cli.py
> index d51df81c..85db06cc 100644
> --- a/tests/test_cli.py
> +++ b/tests/test_cli.py
> @@ -1348,6 +1348,16 @@ c.add_compare(
>      "cloud-init-options5",
>      env={"VIRTINST_TEST_SUITE_PRINT_CLOUDINIT": "1"},
>  )  # --cloud-init user-data=,meta-data=,network-config=
> +c.add_compare(
> +    "--disk %(EXISTIMG1)s --os-variant fedora28 --cloud-init user-data=https://example.com,meta-data=http://example.com,network-config=ftp://foobar.com",
> +    "cloud-init-options6",
> +    env={"VIRTINST_TEST_SUITE_PRINT_CLOUDINIT": "1"},
> +)  # --cloud-init network URLs
> +c.add_invalid(
> +        "--disk none --osinfo fedora28 --cloud-init user-data=badurl://example.com",
> +    env={"VIRTINST_TEST_SUITE_PRINT_CLOUDINIT": "1"},
> +    grep="Couldn't acquire file.*badurl",
> +)  # --cloud-init confirm bad URL fails
>  c.add_valid(
>      "--panic help --disk=? --check=help", grep="path_in_use"
>  )  # Make sure introspection doesn't blow up
> ```

Good point, will add tests before merging.

Issue #870, opened 2025-03-14
virt-xml: Unable to add a vDPA interface due to source.dev


SickMcNugget:
**Describe the bug**
I want to add a NIC manually to my VM config that utilises a vDPA device.
This requires the xml element:
```xml
<source dev="/dev/vhost-vdpa-0"/>
```
virt-xml does not supply the source.dev field in the network category, and complains when I try to use it.
I've tested this with the latest commit #0273621 and with version 4.1.0 of virt-xml.

**To Reproduce**
```bash
virt-xml --network=? | grep source.dev
# There should be no output
```

**Expected behavior**
```bash
virt-xml --network=? | grep source.dev
# outputs: "  source.dev"
```

**System**
 - Distribution: Rocky 9.4
 - virt-manager version: 4.1.0
 - libvirt version: 10.0.0

**Additional context**
This is an example command I used to try and add the nic:
```bash
virt-xml vdpavm --add-device --network type="vdpa",source.dev="/dev/vhost-vdpa-0"
# ERROR    Unknown --network options: ['source.dev']
virt-xml vdpavm --add-device --network type="vdpa",source.path="/dev/vhost-vdpa-0"
# ERROR    XML error: Missing required attribute 'dev' in element 'source'
virt-xml vdpavm --add-device --network type="vdpa"
# ERROR    XML error: interface type='vdpa' requires a 'source' element
```

PR #875, opened 2025-03-18
Add support for vDPA network device

SickMcNugget:
To fix issue #870 

This maps --network source.dev= to the XML:
```xml
<interface type="vdpa">
  <source dev="/dev/vhost-vdpa-0"/>
</interface>
```


phrdina:
Hi @SickMcNugget , thanks for the contribution.

I've reworked your patch a bit to not introduce `source.dev` parameter as we use `source` for other network types so `--network type=vdpa,source=/dev/vhost-vdpa-0` will be the command to use.

I've also added support to virt-manager.

Issue #846, opened 2025-02-18
[feat] Make "USB Host Device" optional by default


sequencerr:
![Image](https://github.com/user-attachments/assets/b720f29a-df62-4f67-a019-32738545ed2c)

![Image](https://github.com/user-attachments/assets/0f63e30b-397b-4907-8cd6-4084a8c8f773)

![Image](https://github.com/user-attachments/assets/9935ee6a-8bf9-49b0-875e-e1a07d30181d)

Make `startupPolicy="optional"` by default and and checkbox in gui for this

morecache:
@phrdina I'd like to take it, May I?

phrdina:
@morecache Sure, take it :) no need to ask. Thanks!

PR #876, opened 2025-03-18
Add startupPolicy for USB passed through devices

morecache:
When the attribute startupPolicy set for a USB passed through devices allows such device to be missing.

The attribute accepts the following values:
* mandatory: fail if missing for any reason (the default)
* requisite: fail if missing on boot up, drop if missing on migrate/restore/revert
* optional:  drop if missing at any start attempt

Fixes: https://github.com/virt-manager/virt-manager/issues/846


phrdina:
I would just squashed the first 3 commits to a single commit, running test suit will otherwise fail for the first two commits.

IMHO the `Add New Virtual Hardware` page looks a bit strange with the `Startup Policy` at the bottom. Not sure how to improve it though.

Otherwise looks good so we just need to figure out the UI design part.

morecache:
v2
How about putting it in the upper right?

phrdina:
Hi, sorry for late reply. Having it on the right side like in that picture doesn't feel right as well. I tried locally several options but eventually ended up with the original placement at the bottom as I was not able to figure out better design.

Cleaned up the ui file a little bit and squashed the first 3 commits into one, thanks for the contribution!

Issue #871, opened 2025-03-15
can't boot 32-bit `debian12` ISO when `--osinfo debian12`


cagney:
Operating system: Fedora 41
Architecture: x86-64
kernel version: 6.13.6-200.fc41.x86_64
libvirt version: libvirt-client-10.6.0-7.fc41.x86_64
Hypervisor and version: qemu-9.1.3-1.fc41.x86_64

While I can boot a 64-bit debian12 ISO with `--osinfo debian12 it doesn't seem to work for 32-bit debian.
The workaround is to specify something like `linix2022` (since the param now seems to be needed)

Steps to reproduce
```
sudo virt-install \
	--connect=qemu:///system \
--check=path_in_use=off 
--graphics=none 
--virt-type=kvm 
--noreboot 
--console=pty,target_type=serial 
--cpu=host-passthrough 
--network=network:swandefault,model=virtio 
--rng=type=random,device=/dev/random 
--security=type=static,model=dac,label='1000:107',relabel=yes 
--vcpus=1 
--memory=2048 
--name=debian-base 
--osinfo debian12 
--disk=path=/home/libreswan/pool/debian-base.qcow2,size=10,bus=virtio,format=qcow2 
--filesystem=target=pool,type=mount,accessmode=squash,source=/home/libreswan/pool 
[...]

PR #878, opened 2025-03-18
Improve Debian ISO detection

phrdina:
Resolves https://github.com/virt-manager/virt-manager/issues/871


crobinso:
Nice one. passes my battery of local iso tests too

PR #888, opened 2025-03-26
Added ability to specify the mode of the Macvtap interface when this is selected. 

farsonic:
MACVTAP has several modes of operation beyond the default of bridge. This change allows the user to specify the mode of operation without having to directly modify the XML in the UI. 

This change has been added to allow users in my case to specifically use mode=private. This restricts communication between devices with the same setting (private) but allows for communication with outside systems, even when on the same VLAN. This can be leveraged with external hardware platforms such as firewalls to facilitate micro-segmentation of traffic on the same VLAN.

### DIFF
--- a/ui/netlist.ui
+++ b/ui/netlist.ui
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- Generated with glade 3.40.0 -->
+<!-- Generated with glade 3.38.2 -->
 <interface>
   <requires lib="gtk+" version="3.22"/>
   <!-- n-columns=2 n-rows=6 -->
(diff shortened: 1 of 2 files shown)


crobinso:
Thank you for the patches, and the suggestion, however not including `mode:` here was a deliberate decision. It was actually removed from UI in jan 2020: https://github.com/virt-manager/virt-manager/commit/227d6982aec639930bec3f00193a35b309012c1a

So I'm closing this, but please read on

libvirt actually has support that lets you get this in a simpler way in virt-manager UI after some initial setup. You can create a virtual network with your desired macvtap settings, which then shows up in the standard network drop down (or name it `default` if you want it as the default network choice). It's documented here: https://libvirt.org/formatnetwork.html#using-a-macvtap-direct-connection

In your case you'd use `<forward dev='your-interface-name' mode='private'/>`

I'd accept patches adding UI for that to virt-managers virtual network creation dialog. Something like:

`Mode:` would get a `MacVTap` option. 
When that is selected:
* All the IP configuration stuff goes away, like is done for SR-IOV pool selection
*  `Forward to:` would select `Physical device...` so user is forced to enter a device
* A new `MacVTap` mode combo box would appear, with bridge, private, etc, defaulting to bridge.
* trustGuestRxFilters checkbox pops up too, defaulting to off, like libvirt does.

Any interest in working on that?

FarsonicScrewdriver:
I might take a look at making that change in the virtual network dialog window.

PR #904, opened 2025-05-06
Add Ctrl+Alt+Shift+Esc key command for logind's SecureAttentionKey

n3rdopolis:
logind now supports a new key binding https://github.com/systemd/systemd/pull/29542 Ctrl+Alt+Shift+Esc that emits SecureAttentionKey to allow login managers to start or switch back to the greeter


ptoscano (approved):
Thanks!

Issue #94, opened 2020-03-27
Setting domain hostname during installation


zippy2:
When installing a guest, it might be desired to set its hostname too. While libosinfo is capable of generating kickstart which sets hostname, this is not exposed in virt-install in any way.

CC @fidencio per our previous IRC discussion.

crobinso:
Probably call it `--unattended hostname=` and someone can follow the pattern of commit a9aeca60bb14c467034170f0c7fb624c57df3fc1

fidencio:
I can see the situation is more complicated than just exposing and it does deserve an issue on osinfo-db as well.

Currrently we'd set the VM name as hostname, cool. Apart that I don't see it happening at all with Fedora (for instance).

So, on the osinfo-db side, we should check for which distros it's failing, have it fixed with the distros, add the attribute as supported or not (also on osinfo-db) and only then expose it here.

fidencio:
This depends on: https://gitlab.com/libosinfo/osinfo-db/-/merge_requests/139

pannadan:
Hi! @thabat12 and I would like to pick up on this issue. We talked to @phrdina about picking up a few CLI issues.

pannadan:
@crobinso can we follow the paradigm from the commit you stated earlier? @thabat12 and I are a little confused on what would complicate this with distros and the osinfo-db that were brought up earlier in the discussion.

PR #906, opened 2025-05-07
hostname support for unattended installs

pannadan:
followed paradigm described in #94 to support setting hostname with unattended installs.

### DIFF
--- a/man/virt-install.rst
+++ b/man/virt-install.rst
@@ -858,6 +858,9 @@ Sub options are:
 ``product-key=``
     Set a Windows product key
 
+``hostname=``
+    Set the hostname for the unattended VM install.
+
 
 
 ``--cloud-init``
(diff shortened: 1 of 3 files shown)


ptoscano (changes requested):
Thanks @pannadan for your PR, and sorry for the late review.

I see that the logic for choosing the hostname of the guest is done in `prepare_install_scripts()`: I don't think this is correct, as similar logics for other unattended options is in `_make_installconfig()`. Hence, what I recommend instead is:
- rename the `hostname` parameter of `_make_installconfig()` from `hostname` to e.g. `guest_name`, to make it clear it is not an hostname
- in `_make_installconfig()` add the logic to either use the provided hostname or use the guest name

Do not forget to:
- add tests for this new scenario
- reformat your changes using the `black` tool
- rebase to the upstream `main` branch

Thanks!

phrdina:
Closing stale PR.

PR #907, opened 2025-05-07
Adding --update-force functionality to override libvirt safety mechanisms

thabat12:

PR #908, opened 2025-05-07
Issue#599: Supporting qemu:override Syntax for libvirt XML Configurations

thabat12:


ptoscano (changes requested):
Hi @thabat12,

thanks for the PR, and sorry for the late answer.

This PR includes also the changes of #907; please drop them from here, so this PR contains only the changes for `qemu:override`.

The new syntax is added as part of the `--qemu-commandline` option; I do not think it is a good idea to conflate the device overriding with the command line handling, so my recommendation is to add a new `--qemu-override` option with its own parser to handle the new syntax.

Please document the new syntax in the man page of `virt-install`, in `man/virt-xml.rst`.

Please add tests for the new syntax, see `tests/test_cli.py`.

Please ensure that the codebase is formatted using the `black` tool -- install it from your distro packages or `pip`, and run it after doing your changes.

Thanks!

phrdina:
Closing stale PR.

Issue #901, opened 2025-04-26
Can't choose a local folder for the storage pool.


djeddi-yacine:
when i want to choose a folder from the storage pool to create a shared file-system, a python error was thrown :
```
UnboundLocalError: cannot access local variable 'start_folder' where it is not associated with a value
Traceback (most recent call last):
  File "/usr/share/virt-manager/virtManager/storagebrowse.py", line 162, in _browse_clicked
    return self._browse_local()
           ~~~~~~~~~~~~~~~~~~^^
  File "/usr/share/virt-manager/virtManager/storagebrowse.py", line 190, in _browse_local
    start_folder=start_folder,
                 ^^^^^^^^^^^^
UnboundLocalError: cannot access local variable 'start_folder' where it is not associated with a value
```
so i edited the lines: 187-191 in `/usr/share/virt-manager/virtManager/storagebrowse.py` :
from :
```py
        filename = self.err.browse_local(
            dialog_type=data.dialog_type,
            dialog_name=data.local_title,
            start_folder=start_folder,
            choose_label=data.choose_label)
```
to :
```py
        filename = self.err.browse_local(
            dialog_type=data.dialog_type,
            dialog_name=data.local_title)
            # start_folder=start_folder,
            # choose_label=data.choose_label)
```
and the problem was solved !!

Eliecer8754:
bro you saved me

phrdina:
Fixed by #911 .

stackrot:
Built from source and confirmed fixed, ty!

PR #911, opened 2025-05-11
fix: set default start_folder value to None

zahidkizmaz:
#901

Error:
```
DEBUG (cli:195) Uncaught exception:
Traceback (most recent call last):
  File "/nix/store/f4iw4ran3xbfxqfcqmc0l69gj33pq3lq-virt-manager-5.0.0/share/virt-manager/virtManager/storagebrowse.py", line 162, in _browse_clicked
    return self._browse_local()
           ^^^^^^^^^^^^^^^^^^^^
  File "/nix/store/f4iw4ran3xbfxqfcqmc0l69gj33pq3lq-virt-manager-5.0.0/share/virt-manager/virtManager/storagebrowse.py", line 190, in _browse_local
    start_folder=start_folder,
                 ^^^^^^^^^^^^
UnboundLocalError: cannot access local variable 'start_folder' where it is not associated with a value
```

### DIFF
--- a/virtManager/storagebrowse.py
+++ b/virtManager/storagebrowse.py
@@ -176,6 +176,7 @@ class vmmStorageBrowser(vmmGObjectUI):
         data = _BrowseReasonMetadata(self._browse_reason)
         gsettings_key = data.gsettings_key
 
+        start_folder = None
         if gsettings_key:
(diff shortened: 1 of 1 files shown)


phrdina:
Hi, thanks for the patch, checking the code this is the only place where start_older would be passed undefined.

PR #922, opened 2025-06-08
data: improve desktop file

fkobi:


ptoscano (changes requested):
Hello @fkobi,

thanks for the PR; there are few things to improve/change/drop:

> sort attributes

By what exactly? They are keys in a INI-like file, and the desktop entry spec does not mandate any order.

fkobi:
> [sorted] By what exactly?

I'd say human readability:
I put `Type` at the top, as it is required and relatively constant, so it can be skimmed over like the `Desktop Entry`.
Then I put the "human readable strings" of Name, GenericName & comment,
After that a mall visual break for Exec & Icon.
At the end there are tags & other specific information.

ptoscano (changes requested):
Thanks for the changes to `Categories` and `Keywords`, they LGTM to me now.

Please revert the sorting/shuffling around of lines, as they do not do any actual change, and only make the git history more complicated.

Thanks!

PR #923, opened 2025-06-11
Various flake8 fixes

ptoscano:
- explicitly exclude E402 in wrapper scripts
- `not .. in/is` -> `not in/is`
- `logging.WARN` -> `logging.WARNING`
- explicit specify flake8 issue in `noqa`
- fix some gettext calls
- merge few strings in the same line that are implicitly concatenated
- drop unneeded `object` inheritance
- drop extra parentheses

See the commit messages for more details.

In most of the commit there are no behaviour changes. The "gettext calls fix" commit actually fixes the translation of few strings.

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -2686,7 +2686,7 @@ def test_virtinstall_no_testsuite():
     some code paths like proper logging etc.
     """
     cmd = Command(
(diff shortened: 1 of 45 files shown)


phrdina:
Thanks, nice cleanup!

Issue #897, opened 2025-04-13
libxml2's Python bindings are being deprecated


nwellnhof:
`virt-manager` is one of the very few open-source projects still using libxml2's original Python bindings. The libxml2 bindings are outdated and insecure and are considered for removal in the long term.

nambona890:
The bindings are deprecated to the point of me being unable to run virt-manager on my Arch Linux system as I can't install libxml2-python3 through pip, due to numerous errors and warnings when trying to build.
Some notes are that I also have lxml installed in pip, and that I added a home directory Python venv ("~/.pyenv/bin") to $PATH.
```
Collecting libxml2-python3
  Using cached libxml2-python3-2.9.5.tar.gz (216 kB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
Building wheels for collected packages: libxml2-python3
  Building wheel for libxml2-python3 (pyproject.toml) ... error
  error: subprocess-exited-with-error
  
  × Building wheel for libxml2-python3 (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [179 lines of output]
      In file included from libxml_wrap.h:7,
                       from libxml.c:28:
      /usr/include/libxml2/libxml/nanoftp.h:13:4: warning: #warning "libxml/nanoftp.h is deprecated" [-Wcpp]
         13 |   #warning "libxml/nanoftp.h is deprecated"
            |    ^~~~~~~
      In file included from libxml_wrap.h:17:
      /usr/include/libxml2/libxml/xmlunicode.h:12:4: warning: #warning "libxml/xmlunicode.h is deprecated" [-Wcpp]
         12 |   #warning "libxml/xmlunicode.h is deprecated"
            |    ^~~~~~~
      libxml.c: In function ‘libxml_xmlDumpMemory’:
      libxml.c:231:9: warning: ‘xmlMemoryDump’ is deprecated [-Wdeprecated-declarations]
        231 |         xmlMemoryDump();
            |         ^~~~~~~~~~~~~
      In file included from libxml.c:17:
      /usr/include/libxml2/libxml/xmlmemory.h:161:9: note: declared here
        161 |         xmlMemoryDump   (void);
            |         ^~~~~~~~~~~~~
      libxml.c: In function ‘xmlPythonFileCloseRaw’:
[...]

PR #927, opened 2025-06-18
xmlapi: rewrite the code to use lxml instead of libxml2

phrdina:
The python module libxml2 is part of the libxml2 library and they are planning to deprecate and remove this binding in favor of lxml that also uses libxml2 but the binding is done in more pythonic way.

The change to use lxml uncovered some incorrectly formatted bits in our test XMLs that are now fixed.

Disable new pylint category c-extension-no-member as lxml is not introspected by default and pylint complains about it.

Resolves: https://github.com/virt-manager/virt-manager/issues/897

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -3,7 +3,7 @@
 # can either give multiple identifier separated by comma (,) or put this option
 # multiple time (only on the command line, not in the configuration file where
 # it should appear only once).
-disable=Design,Similarities,invalid-name,missing-docstring,line-too-long,too-many-lines,superfluous-parens,bad-whitespace,locally-disabled,no-self-use,unnecessary-lambda,star-args,fixme,global-statement,bare-except,anomalous-backslash-in-string,broad-except,cyclic-import,bad-continuation,locally-enabled,unidiomatic-typecheck,redefined-variable-type,bad-option-value,wrong-import-position,consider-using-ternary,no-else-return,len-as-condition,inconsistent-return-statements,useless-object-inheritance,consider-using-in,useless-return,assignment-from-none,chained-comparison,import-outside-toplevel,use-a-generator,consider-using-with,consider-using-f-string,unspecified-encoding,use-maxsplit-arg,possibly-used-before-assignment
+disable=Design,Similarities,invalid-name,missing-docstring,line-too-long,too-many-lines,superfluous-parens,bad-whitespace,locally-disabled,no-self-use,unnecessary-lambda,star-args,fixme,global-statement,bare-except,anomalous-backslash-in-string,broad-except,cyclic-import,bad-continuation,locally-enabled,unidiomatic-typecheck,redefined-variable-type,bad-option-value,wrong-import-position,consider-using-ternary,no-else-return,len-as-condition,inconsistent-return-statements,useless-object-inheritance,consider-using-in,useless-return,assignment-from-none,chained-comparison,import-outside-toplevel,use-a-generator,consider-using-with,consider-using-f-string,unspecified-encoding,use-maxsplit-arg,possibly-used-before-assignment,c-extension-no-member
 enable=fixme
 
 
--- a/tests/data/cli/compare/virt-xml-update-nodefine-succeed.xml
+++ b/tests/data/cli/compare/virt-xml-update-nodefine-succeed.xml
@@ -11,10 +11,10 @@
    </devices>
  </domain>
 
-    <interface type="user">
-      <mac address="00:11:22:33:44:56"/>
-      <model type="e1000"/>
-    </interface>
+<interface type="user">
+  <mac address="00:11:22:33:44:56"/>
+  <model type="e1000"/>
+</interface>
 
 
 Hotplug this device to the guest 'test'? (y/n): Device hotplug successful.
--- a/tests/data/cli/compare/virt-xml-update-succeed.xml
+++ b/tests/data/cli/compare/virt-xml-update-succeed.xml
@@ -11,10 +11,10 @@
    </devices>
  </domain>
 
-    <interface type="user">
-      <mac address="00:11:22:33:44:55"/>
-      <model type="e1000"/>
-    </interface>
+<interface type="user">
+  <mac address="00:11:22:33:44:55"/>
+  <model type="e1000"/>
+</interface>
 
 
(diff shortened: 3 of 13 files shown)


ptoscano:
Thanks @phrdina for the changes -- it's nice to see that, except from whitespace changes in generated XMLs, the swapping of XML implementation causes no changes elsewhere.

> Disable new pylint category c-extension-no-member as lxml is not introspected by default and pylint complains about it.

Rather than adding it in `.pylintrc`, I'd add it as inline exclusion for the whole `virtinst/xmlapi.py` only, in the same way done in line 12:
```python
# pylint: disable=protected-access
```
this way it applies only to this file.

Regarding the very small loss of coverage: I see the helper `node_is_text()` in `virtinst.xmlapi` is reported as not covered. This seems correct to me, and it is due to the fact that it is not used anymore. So simply drop it altogether.

phrdina:
> Thanks @phrdina for the changes -- it's nice to see that, except from whitespace changes in generated XMLs, the swapping of XML implementation causes no changes elsewhere.
> 
> > Disable new pylint category c-extension-no-member as lxml is not introspected by default and pylint complains about it.
> 
> Rather than adding it in `.pylintrc`, I'd add it as inline exclusion for the whole `virtinst/xmlapi.py` only, in the same way done in line 12:

Good point, will do, thanks!

> Regarding the very small loss of coverage: I see the helper `node_is_text()` in `virtinst.xmlapi` is reported as not covered. This seems correct to me, and it is due to the fact that it is not used anymore. So simply drop it altogether.

Nice catch, I missed that one and will remove it, there are other places that are no longer covered by CLI tests but are still used by GUI. I'll look into it if there is some easy fix for it.

crobinso:
Sorry just seeing this now. TLDR I hope this works, would love to drop libxml2 bindings, but we need to investigate more since lxml usage has caused crashes with libvirt in the past.


I looked at this a long while ago: https://github.com/crobinso/virt-manager/commits/xml-etree/

But at the time, lxml needed to use global libxml2 hooks to do its thing. Via pip, where libxml2 is staticly linked into lxml, that's fine. But with distro packages, where lxml and libvirt both use libxml2 shared library, the hooks lxml installed interfered with libvirt and caused segfaults. 

There's some more info here:
https://bugzilla.redhat.com/show_bug.cgi?id=1544019
https://bugs.launchpad.net/lxml/+bug/1748019

That launchpad bug mentions this commit that improved things in 2020:
https://github.com/lxml/lxml/commit/fa1d856cad369d0ac64323ddec14b02281491706

Which is basically what I intially suggested. But looking now I don't think that's actually safe for multithreading use. From latest libxml2 source code

```c
/*                                                                               
 * This global has to die eventually                                             
 */                                                                              
static xmlExternalEntityLoader                                                   
xmlCurrentExternalEntityLoader = xmlDefaultExternalEntityLoader;                 
                                                                                 
/**                                                                              
 * Changes the default external entity resolver function for the                 
 * application.                                                                  
 *                                                                               
 * @deprecated This is a global setting and not thread-safe. Use                 
 * #xmlCtxtSetResourceLoader or similar functions.                               
 *                                                                               
 * @param f  the new entity resolver function                                    
 */                                                                              
void                                                                             
xmlSetExternalEntityLoader(xmlExternalEntityLoader f) {                          
    xmlCurrentExternalEntityLoader = f;                                          
} 
```

So those global functions aren't thread safe, and are deprecated, and I don't see any reference to SetResourceLoader in current lxml code. So I'm nervous here. We need to investigate more

crobinso:
I'll also mention my old branch has minidom and stock ElementTree implementations for xmlapi, but both require hacks to not needlessly reorder XML: https://github.com/crobinso/virt-manager/commit/b7c553638b033258c4e9f5aac42671c390db85e0

berrange:
> So those global functions aren't thread safe, and are deprecated, and I don't see any reference to SetResourceLoader in current lxml code. So I'm nervous here. We need to investigate more

Yeah, that code in lxml looks decidedly unsafe. It may be just good in a pure python program, but as soon as C level threads are involved that's inevitably going to fail on us. It needs to be ported to use xmlCtxtSetResourceLoader, but it makes me wonder if there's other use of global vars in lxml that may trip us up, as libxml2 has loads of historical global state that shouldn't be touched.



> I'll also mention my old branch has minidom and stock ElementTree implementations for xmlapi, but both require hacks to not needlessly reorder XML: [crobinso@b7c5536](https://github.com/crobinso/virt-manager/commit/b7c553638b033258c4e9f5aac42671c390db85e0)

The test suite reordering doesn't sound too bad - we don't need to churn the test files - just roundtrip our reference xml files through minidom, so that they get re-ordered to match the ordering on the test output. 

The diffs on virt-xml, however, are a pretty big concern, changes to XML by virt-xml needs to be only touching the specific bits being changed.

berrange:
NB, libxml  2.15.0 has now disabled its python bindings by default, and 2.16.0 will finally delete them. So a solution is getting pretty urgent.

crobinso:
I'll see about dusting off my old plain ElementTree/minidom branch

crobinso:
My branch here is passing the test suite now: https://github.com/crobinso/virt-manager/tree/xml-etree2

Need to use `VIRTINST_XML_BACKEND=etree` env variable to enable it.
Test suite goes from 24s to 42s

There's a lot of hacking around default python behavior to get the output to behave. 
It's less sketchy than the hacks I used in the old branch, but it still aint pretty. It will definitely cause maintenance pain at some point.
For our own sanity we need to look into fixing lxml itself. If no one raises their hand I will try to carve out time this month to explore it.

Still todo for my code:
* clean it up, commit messages, pylint, check code coverage, etc.
* test with multiple python versions. we probably need to bump minimum python version to at least 3.8
* test it with round tripping every XML file in libvirt's test suite.

Note, i'd be fine if we kept etree, libxml2, and lxml backend impls in tree for a while until we have a clear path forward.

crobinso:
Ok I submitted a PR with a stock etree implementation: https://github.com/virt-manager/virt-manager/pull/983

I looked into lxml some more but I need to write my findings. I'll put them in the libxml2 issue: https://github.com/virt-manager/virt-manager/issues/897

crobinso:
I filed an issue about moving to lxml and put all my thoughts there: https://github.com/virt-manager/virt-manager/issues/984

PR #931, opened 2025-06-19
tests: pytest-related improvements

ptoscano:
- use `pytest.raises()` instead of `try`/`except` blocks
- `test_urls.py`: make use of parametrized tests instead of manual generation
- `test_checkprops.py`: use directly `pytest.fail()` rather than a `try`+`raise`+`except` game

See the messages of each commit for more details. There should be no behaviour changes to the tests.


phrdina:
Nice cleanup, thanks!

PR #933, opened 2025-06-20
RFC: enable ruff as linter

ptoscano:
ruff [1] is a modern and extremely fast Python linter, which replaces e.g. flake8 and provides OOTB additional flake8 checks. 

ruff cannot yet fully replace pylint (see [2]), so for now pylint is kept.

Configure ruff (in `pyproject.toml`) so it can be usable with the current
codebase; in particular:
- borrow some of the existing configuration knobs from black and pylint, e.g. the line length, the embedded functions, the dummy variables regex
- enable various linters that apply to the codebase (e.g. not the ones for typing, or async, pandas, etc), and I deem uncontroversial enough to be used
   - see [3] for the list of existing rules; I'm definitely open for discussion about what should be enabled by default
- add various per-file ignores according to what the existing files trigger; this is not meant to be a "set and forget", but rather a way to record the existing issues per-file and avoid those issues to show up in new files

[1] https://docs.astral.sh/ruff/
[2] https://github.com/astral-sh/ruff/issues/970
[3] https://docs.astral.sh/ruff/rules/


ptoscano:
This depends on #932 for fixing/suppressing most of the E501 issues.

Issue #785, opened 2024-12-07
Test suite since conversion to meson build


Toolybird:
**Distro**: Arch Linux

Hi, I've been building `virt-manager` on my Arch system for many years. I used to be able to run the test suite and have it happily pass (most of the time). But things seem to have degraded since the conversion to `meson`. This is what I'm currently getting:
```
1/3 pycodestyle OK               4.42s
2/3 pytest      OK              24.44s
3/3 pylint      FAIL            52.81s   exit status 6
>>> MALLOC_PERTURB_=0 MSAN_OPTIONS=halt_on_error=1:abort_on_error=1:print_summary=1:print_stacktrace=1 MESON_TEST_ITERATION=1 ASAN_OPTIONS=halt_on_error=1:abort_on_error=1:print_summary=1 UBSAN_OPTIONS=halt_on_error=1:abort_on_error=1:print_summary=1:print_stacktrace=1 /usr/bin/pylint tests virtinst virtManager --rcfile .pylintrc
――――――――――――――――――――――――――――――――――――― ✀  ―――――――――――――――――――――――――――――――――――――
stdout:
************* Module tests.uitests.lib._node
tests/uitests/lib/_node.py:8:0: E0401: Unable to import 'dogtail.tree' (import-error)
tests/uitests/lib/_node.py:9:0: E0401: Unable to import 'pyatspi' (import-error)
tests/uitests/lib/_node.py:253:8: W0201: Attribute 'text' defined outside __init__ (attribute-defined-outside-init)
************* Module tests.uitests.lib.app
tests/uitests/lib/app.py:11:0: E0401: Unable to import 'dogtail.rawinput' (import-error)
tests/uitests/lib/app.py:12:0: E0401: Unable to import 'dogtail.tree' (import-error)
************* Module tests.uitests.lib.utils
tests/uitests/lib/utils.py:6:0: E0401: Unable to import 'dogtail.config' (import-error)
************* Module tests.uitests.lib._dogtailinit
tests/uitests/lib/_dogtailinit.py:16:0: E0401: Unable to import 'dogtail.config' (import-error)
tests/uitests/lib/_dogtailinit.py:17:0: E0401: Unable to import 'dogtail.utils' (import-error)
tests/uitests/lib/_dogtailinit.py:48:0: E0401: Unable to import 'dogtail.tree' (import-error)
************* Module virtinst.cli
virtinst/cli.py:541:4: E0401: Unable to import 'argcomplete' (import-error)
stderr:
[...]

PR #937, opened 2025-06-25
Run pylint only as GitHub workflow

ptoscano:
pylint is more a static analyzer than a test suite, so it does not fit too well when running the tests. In addition to that, it requires additional Python modules needed only when running the UI tests: https://github.com/virt-manager/virt-manager/issues/785

Hence, run pylint as additional github workflow. This actually makes pylint run for PRs, rather than relying on maintainers and contributors to remember to run it manually.

pylint is added as github workflow rather than additional pre-commit hook, as the pylint documentation discourages the pre-commit integration:
https://pylint.readthedocs.io/en/stable/user_guide/installation/pre-commit-integration.html


phrdina:
I'm not sure if we should remove pylint from meson. Developers that would like to run pylint locally would have to manually invoke `pylint tests virtinst virtManager`.

I wanted to suggest to use different suite and skip it by default but it seems like meson doesn't allow skipping tests. We can still set `suite` for each test to group them and developers/users can run `meson -C build test --suite XXX` or `meson -C build test --no-suite YYY`.

In addition but that is for different PR we should make it easy to run every single tool we have in gitlab workflow locally.

ptoscano:
> In addition but that is for different PR we should make it easy to run every single tool we have in gitlab workflow locally.

My goal is to leverage pre-commit to run linters and checkers for this as much as possible. My attempt to add codespell to it failed (#940) though, I'll give it a try again in the future.

Sadly pylint does not play well with pre-commit (see the link above), otherwise I'd have added pylint there.

phrdina:
> > In addition but that is for different PR we should make it easy to run every single tool we have in gitlab workflow locally.
> 
> My goal is to leverage pre-commit to run linters and checkers for this as much as possible. My attempt to add codespell to it failed (#940) though, I'll give it a try again in the future.
> 
> Sadly pylint does not play well with pre-commit (see the link above), otherwise I'd have added pylint there.

Using pre-commit hook is nice addition but it should not be the only way how to run linters locally. For example I'm not willing to install that project as my pre-commit hook as I don't trust it and I don't want to verify every single thing it downloads from internet.

PR #940, opened 2025-06-27
ci: run codespell as pre-commit hook

ptoscano:
This makes the github workflow for it obsolete, as pre-commit is run already as github workflow.


ptoscano:
Hm it does not seem to keep into account the `skip` bits in `.codespellrc`; I'll close this for now, since the github workflow works fine anyway.

PR #943, opened 2025-06-30
tests: use more pytest 7+ pathlib bits

ptoscano:
pytest 7 adds various bits using `pathlib`, deprecating the old counterparts based on `py.path`; in particular, in virt-manager I found two places:
- the `pytest_ignore_collect()` hook: there is a new `collection_path` parameter replacing `path`, however there is no way to keep a single implementation, extra steps are done to implement the right hook depending on the pytest version
- `Node.fspath`, in particular as used in the `pytest_collection_modifyitems()` hook: add a new helper to use the new `Node.path` when available

See the messages of the commits for longer explanations. This should make it possible to run the tests without any behaviour change in any pytest version.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -87,8 +87,14 @@ else:
 
 
 def pytest_collection_modifyitems(config, items):
+    def item_path_basename(item):
+        try:
+            return item.path.name
+        except AttributeError:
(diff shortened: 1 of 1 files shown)


ptoscano:
Sigh, pushed to the main repository instead of my fork... #944

PR #948, opened 2025-07-09
virtinst: add support for creating TDX guests

berrange:
A minimal config to enable TDX, with support for attestation would be

 $ virt-install
     ...args...
     --boot uefi \
     --machine q35 \
     --launchSecurity type=tdx,quoteGenerationService=on

NB, libvirt support for TDX is NOT yet merged, and this PR XML does NOT patch the XML of the latest patches - it is written based on my suggested XML rename of one of the elements. 

IOW this PR must wait until the final libvirt merge.

### DIFF
--- a/tests/data/cli/compare/virt-install-x86_64-launch-security-tdx-full.xml
+++ b/tests/data/cli/compare/virt-install-x86_64-launch-security-tdx-full.xml
@@ -1,0 +1,83 @@
+<domain type="kvm">
+  <name>vm1</name>
+  <uuid>00000000-1111-2222-3333-444444444444</uuid>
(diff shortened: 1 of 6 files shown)


berrange:
The libvirt TDX patches were merged today and will be in the 11.6.0 release next Friday

berrange:
The codecov complaints are presumably due to the new unit tests being conditionalized on 11.6.0, which is not available in the test environment.

phrdina:
> The codecov complaints are presumably due to the new unit tests being conditionalized on 11.6.0, which is not available in the test environment.

Yeah most likely, I'll look at it on Monday.

PR #957, opened 2025-07-30
addhardware: network: adjust supported network devices for bhyve

novel:
- Add bhyve to a list of guest types that support virtio device type
 - Do not include rtl8139 for bhyve


novel:
ping?

PR #983, opened 2025-09-23
Add XML backend using stock python ElementTree

crobinso:
This is an XMLAPI backend using stock python ElementTree.
We need to extend and re-implement some of ElementTree internals
to make its output match what libvirt generates, so virt-xml edits
don't generate extraneous diffs.
    
This is disabled by default but will be used if libxml2 is not
installed. You can explicitly opt in to using it by setting
env var VIRTINST_XML_BACKEND=etree before virtinst is imported.

Tested with fedora 42 with python 3.13, and rhel9 with python 3.9
I also roundtripped all the domain XML in the libvirt test suite using the new backend.

Fixes: #897


crobinso:
some other bits:

* I didn't attempt to test with anything older, but less than python 3.8 probably won't work due to some of the ElementTree features we are using. so maybe we should bump minimum python version (currently we check for 3.4)

* @phrdina you could adapt your lxml series to follow this pattern, putting impl in xmllxml.py and can opt in with VIRTINST_XML_BACKEND=lxml . I don't know if it will be easy or messy to share etree impls between stock python and lxml, I'd be fine with separate impls if you think it's best.

phrdina:
Overall looks good and if we assume same support matrix as libvirt project python 3.9 should be the oldest we need to aim for.

Will definitely adapt the lxml changes to follow this pattern.

I'll look at the ElementTree implementation on Monday, but at the first glance it looks good and solves the libxml2 deprecation issue until lxml is usable.

Thanks!

novel:
Thanks for addressing that. Are there plans to make a new release in the near future?

PR #985, opened 2025-09-26
Add minimize button to the overlay toolbar

gdarko:
This PR introduces a Minimize button to the fullscreen overlay toolbar so you can minimize the window straight from fullscreen, instead of exiting fullscreen and then clicking the window’s minimize control.

image


gdarko:
Hey @phrdina - 

Sorry to be pain. 

I found this change along with #995  very useful - so would like to get those checked, i think it's very beneficial to the project. :-)

Those PRs will add toolbar actions that make the navigation/switching between full-screen VMs breeze, something that VMWare Workstation does really good and this project lacks. I am running a custom build of virt-manager with those changes for months and it works really well.

Thank you for your time!

phrdina:
@gdarko Hi, sorry for taking long time to review this PR. The changes look good but some cleanup is required:

- please make sure there is only single commit in the PR and make sure the commit message contains description of the change, it's good to add description in PR but we also need it in the commit as well, you can search for `git commit best practices` for details how good commit should look like
- make sure you run `black` tool to properly format the code

phrdina:
We will need one modification to the code, I tried running it with sway windows manager and the minimize action is ignored and that might be true for other window manager. I did some research and GTK can provide recommended buttons using `gtk-decoration-layout` setting, more details here https://docs.gtk.org/gtk3/property.Settings.gtk-decoration-layout.html.

We should check if `minimize` is part of the returned string and show the minimize button only in that case.

PR #988, opened 2025-10-03
Rename AppStream metadata to rDNS format

AsciiWolf:
Use proper reverse DNS format for AppStream metadata id as recommended by the spec. Also change the suffix to metainfo.xml and component type to desktop-application.

### DIFF
--- a/data/meson.build
+++ b/data/meson.build
@@ -2,7 +2,7 @@ fs = import('fs')
 
 desktop_file = files('virt-manager.desktop.in')
 
-appdata_file = files('virt-manager.appdata.xml.in')
+metainfo_file = files('org.virt_manager.virt-manager.metainfo.xml.in')
 
 install_data(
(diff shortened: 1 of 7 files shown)


AsciiWolf:
/cc @ptoscano @phrdina

If you have any questions, feel free to ask. Thanks!

ptoscano:
There's a CI failure (code formatting using `black`).

Regarding the AppStream metadata rename: in case `<id>` is renamed, then the old ID must be added as `<provides>`, otherwise old references to it break; please check https://freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-provides

AsciiWolf:
It is actually not needed and as far as I know, the app id rename will only break old user reviews in ODRS, but fair point. Added.

phrdina:
Looks good, one more detail, please wrap the commit message body lines to around 72 characters.

AsciiWolf:
Done. :)

PR #995, opened 2025-10-18
Add VM Switcher to the overlay toolbar

gdarko:
As part of my ongoing ease-of-use improvements for the Fullscreen Overlay Toolbar, this pull request introduces a "VM Switcher" dropdown. This work follows my previous contribution that added a minimize button to the same toolbar (PR #985).

The new dropdown addresses a common pain point for users running multiple virtual machines. Previously, switching between consoles required exiting fullscreen mode. With this change, users can now seamlessly switch between any open VM console directly from the toolbar, significantly improving efficiency.

image

 ### What's Added:

  **VM Switcher Button**
  - New button in the fullscreen toolbar (after "Leave Fullscreen" and "Send Key")
  - Click it to see a list of all VM windows on your current monitor
  - Current VM is marked with a checkmark and grayed out
  - Click any other VM to switch to it immediately

  **Screen-Aware Filtering**
  - Only shows VMs that are on the same monitor as your current fullscreen window
  - If you have VMs spread across multiple monitors, you'll only see the ones on the current screen
  - Prevents clutter and confusion when working with multi-monitor setups

  **Always Available**
  - The button is always visible in the toolbar
  - Menu updates dynamically - closed windows disappear automatically
  - Works with any number of VMs (shows "No other VM windows open" when you only have one)

  ### How It Works

  When you click the VM Switcher button:
  1. Menu populates with all VM windows on the current screen
  2. Current VM has a checkmark and is disabled (can't switch to yourself)
  3. Click any other VM to bring its window to the front

  The switcher uses GTK's monitor detection to figure out which screen each window is on, so it's smart about multi-monitor setups.

### Tests

As a part of this PR i added test in tests/uitests/test_livetests.py:

  - New testConsoleVMSwitcher() test that:
    - Starts two VMs
    - Opens both console windows
    - Enters fullscreen and reveals toolbar
    - Verifies switcher button appears and menu works correctly
    - Tests switching between windows
    - Cleans up

### DIFF
--- a/tests/uitests/test_livetests.py
+++ b/tests/uitests/test_livetests.py
@@ -654,3 +654,151 @@ def testFirmwareRename(app, dom):
     # Confirm nvram paths were altered as expected
     assert not DeviceDisk.path_definitely_exists(app.conn, origpath)
     assert DeviceDisk.path_definitely_exists(app.conn, newpath)
+
+
+def testConsoleVMSwitcher(app):
+    """
+    Test the VM Switcher button in the fullscreen overlay toolbar
+    """
+    # Start two VMs for testing the switcher
+    app.error_if_already_running()
+
+    # Define and start first VM
+    xmlfile1 = "%s/live/uitests-vnc-standard.xml" % tests.utils.UITESTDATADIR
+    conn = libvirt.open("qemu:///system")
+    dom1 = conn.defineXML(open(xmlfile1).read())
+    dom1.create()
+
+    # Define and start second VM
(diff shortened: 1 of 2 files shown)


phrdina:
Hi, sorry that it took me long time to get to this PR. I don't think this is something we should add to virt-manager. I don't see a large user-base for this feature.

There are also technical challenges, it may work in your specific setup but I see at least one issue with multiple workspaces. The code only detect monitors, but for example I usually have multiple workspaces on a single monitor and currently it would list all VMs from all workspaces for the current monitor.

Per our [DESIGN.md](https://github.com/virt-manager/virt-manager/blob/main/DESIGN.md) document I'm closing this PR.

PR #997, opened 2025-10-24
virtinst: add support for acpi-generic-initiator

NathanChenNVIDIA:
A minimal config to enable acpi-generic-initiator would be

 $ virt-install
     ...args...
     --hostdev net_00_1c_25_10_b1_e4,acpi_nodeset=0,2-4

Signed-off-by: Nathan Chen <nathanc@nvidia.com>


phrdina:
I've removed the alias and updated tests, thanks!

PR #1017, opened 2025-12-06
Updated Portuguese Portugal translation

rottenpants466:
Updated the Portuguese Portugal translation.

Changed the .po file to pt_PT because there are some areas in Virtual Manager that appear in English, like you can see here:

Screenshot_20251206_120040

Screenshot_20251206_120107

Screenshot_20251206_120223

A similar issue occurred in these applications: 
https://github.com/alainm23/planify/pull/2049
https://github.com/rafatosta/zapzap/issues/466

By changing pt.po to pt_PT.po it solved the problem. The same should probably be done in weblate too.

Running the locale command gives this result:
```
[andre@Andre-PC ~]$ locale
LANG=pt_PT.UTF-8
LC_CTYPE="pt_PT.UTF-8"
LC_NUMERIC=pt_PT.UTF-8
LC_TIME=pt_PT.UTF-8
LC_COLLATE="pt_PT.UTF-8"
LC_MONETARY=pt_PT.UTF-8
LC_MESSAGES="pt_PT.UTF-8"
LC_PAPER=pt_PT.UTF-8
LC_NAME=pt_PT.UTF-8
LC_ADDRESS=pt_PT.UTF-8
LC_TELEPHONE=pt_PT.UTF-8
LC_MEASUREMENT=pt_PT.UTF-8
LC_IDENTIFICATION=pt_PT.UTF-8
LC_ALL=
[andre@Andre-PC ~]$ 
```

So the expected language is pt_PT not pt.

cheers

### DIFF
--- a/po/pt_PT.po
+++ b/po/pt_PT.po
@@ -1,4 +1,3 @@
-# SOME DESCRIPTIVE TITLE.
 # Copyright (C) YEAR THE PACKAGE'S COPYRIGHT HOLDER
 # This file is distributed under the same license as the PACKAGE package.
 #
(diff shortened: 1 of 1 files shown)


phrdina:
Hi, I don't think this change is correct. If I run `LC_ALL=pt_PT.utf8 virt-manager` on my system I get most of the strings translated. From my understanding using pt.po should work as a fallback for any pt_XX locale if there is not specific locale like it is for pt_BR.

rottenpants466:
> Hi, I don't think this change is correct. If I run `LC_ALL=pt_PT.utf8 virt-manager` on my system I get most of the strings translated. From my understanding using pt.po should work as a fallback for any pt_XX locale if there is not specific locale like it is for pt_BR.

I'm using the flatpak version of Virtual Manager, so I don't know if it's related to that.

phrdina:
Could be related to this issue https://github.com/virt-manager/virt-manager/issues/1023 that was fixed by this commit https://github.com/jmpoep/virt-manager/commit/d13271422e47d9bc827d6ede3e7d5154568115c7 .

rottenpants466:
> Could be related to this issue #1023 that was fixed by this commit [jmpoep@d132714](https://github.com/jmpoep/virt-manager/commit/d13271422e47d9bc827d6ede3e7d5154568115c7) .

Hope it fixes the issue in the next version of Virtual Manager. If it does then there's no need to rename pt.po to pt_PT.po.

PR #1018, opened 2025-12-09
virtinst: Fix XDG_DATA_HOME handling

gmbr3:
The default value of XDG_DATA_HOME should be ~/.local/share.

It will not by default contain libvirt at the end of the path, so subsequently append libvirt after retrieving the value of XDG_DATA_HOME


phrdina:
Thanks for the patch!

PR #1032, opened 2026-01-21
add custom model for loongarch64

DWwanghao:
I encountered missing default configurations and machine types while testing the virtual machine using the official latest ISO on Alpine. Therefore, I have made the following additions. The relevant test cases have also been successfully executed and updated, resulting in a perfect operational outcome as shown below:
Screenshot From 2026-01-21 10-28-44
Screenshot From 2026-01-21 10-30-18

Official Alpine ISO Image:
https://dl-cdn.alpinelinux.org/alpine/v3.23/releases/loongarch64/alpine-standard-3.23.0-loongarch64.iso

The ISO provided above works without issues on both the 3A6000 and 3C5000 models. However, there may be minor issues on other LoongArch physical machines. Alternatively, you can obtain the latest openEuler operating system that supports the LoongArch architecture via the following link:
https://atomgit.com/src-openeuler/virt-manager/tree/openEuler-24.03-LTS-SP3

### DIFF
--- a/virtinst/devices/disk.py
+++ b/virtinst/devices/disk.py
@@ -1049,6 +1049,8 @@ class DeviceDisk(Device):
             return "virtio"
         if guest.os.is_q35():
             return "sata"
+         if self.is_cdrom() and guest.os.is_loongarch64():
+            return "scsi"
         if self.conn.is_bhyve():
             # IDE bus is not supported by bhyve
             return "sata"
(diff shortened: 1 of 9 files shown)


virtinst/devices/disk.py
```diff
@@ -1049,6 +1049,8 @@ def _default_bus(self, guest):
             return "virtio"
         if guest.os.is_q35():
             return "sata"
+         if self.is_cdrom() and guest.os.is_loongarch64():
+            return "scsi"
```
phrdina: This is not needed, few lines below we call `if self.is_cdrom() and guest.supports_virtioscsi():` which returns true for Alpine Linux 3.11 and newer.

phrdina:
Closing this PR as the changes are not necessary.

PR #1034, opened 2026-01-29
virtinst: Prefer NVMe over IDE, SATA, SD and USB

akihikodaki:
NVMe is a modern and more efficient alternative to IDE, SATA, SD and USB.

It is explicitly checked that both the host and guest support the bus because its support is relatively limited compared to the legacy buses.

It is only compatible with disk (i.e., not floppy or CD-ROM).


akihikodaki:
devices: Add NVM Express Revision 1.0 (!888) · Merge requests · libosinfo / osinfo-db · GitLab
https://gitlab.com/libosinfo/osinfo-db/-/merge_requests/888

Issue #495, opened 2023-02-17
RFE: virt-install should add secure-boot convenience options


andreabolognani:
Right now, enabling and disabling Secure Boot requires fairly unwieldy incantations such as `--boot uefi,firmware.feature0.name=secure-boot,firmware.feature0.enabled=yes,firmware.feature1.name=enrolled-keys,firmware.feature1.enabled=yes`.

It would be great if virt-install implemented more user-friendly shorthands for that, such as `--boot uefi,firmware.secure-boot=yes`.

Where things get hairy is that you have two ways to disable Secure Boot: get libvirt to pick either a firmware that contains the feature but doesn't have certificates enrolled, or one where the feature is absent altogether. Depending on the firmware shipped with your distro, you might want one or the other. So maybe a sort of tristate would be more fitting, with the accepted option being `enabled`, `disabled` and `absent`.

Ideally this would also be wired up in the GUI: firmware type (UEFI/BIOS) is already one of the choices that you're presented with, and having one more knob for the UEFI case doesn't seem unreasonable.

Another idea that's floating around is the possibility of enrolling specific Secure Boot certificates at VM creation time. That could be handled by a second key, along the lines of `secure-boot-enroll=microsoft`, `secure-boot-enroll=redhat`, `secure-boot-enroll=path/to/custom/certs`. Probably too specialized to be exposed via the GUI.

andreabolognani:
It could also be useful to have an option in `virt-xml` that can be used to flip the status of Secure Boot after the VM has been installed.

crobinso:
@andreabolognani so I just took a look at this, but fedora x86 doesn't seem to ship any non-secureboot ovmf builds? So trying to toggle secureboot `off` via the features XML just throws errors: `Unable to find 'efi' firmware that is compatible with the current configuration
Failed.`

Not necessarily a blocker to adding convenience options to virt-install, but makes me wonder if world has changed since this was filed and toggling secureboot off is now less practical...

Another case that this would be useful is for toggling on the firmware.json `verbose-static` feature to get debug output out of aarch64 boots, but it doesn't look like libvirt supports that in the feature list?

phrdina:
[...]

PR #1038, opened 2026-02-06
cli: add --boot secure-boot option

phrdina:
The new option can be used to enable/disable secure boot verification of UEFI firmware.

If virt-xml is used to change secure-boot print warning that resetting NVRAM is required to make the change effective.

Fixes: https://github.com/virt-manager/virt-manager/issues/495

### DIFF
--- a/man/virt-install.rst
+++ b/man/virt-install.rst
@@ -973,6 +973,12 @@ Some examples:
 ``--boot uefi=off``
     Do not use UEFI if the VM would normally default to it.
 
+``--boot uefi=on,secure-boot=off``
+    Configure the VM to boot from UEFI with secure-boot enabled and enforced.
+    This requires libvirt with firmware auto-selection. Setting ``secure-boot``
+    to off ensures the firmware can boot unsigned binaries.
+    This is a convenience option to control the enrolled-keys firmware feature.
+
 ``--boot uefi,firmware.feature0.name=secure-boot,firmware.feature0.enabled=yes,firmware.feature1.name=enrolled-keys,firmware.feature1.enabled=yes``
     Configure the VM to boot from UEFI with Secure Boot support enabled.
     Only signed operating systems will be able to boot with this configuration.
--- a/tests/data/cli/compare/virt-install-boot-uefi-secure-boot.xml
+++ b/tests/data/cli/compare/virt-install-boot-uefi-secure-boot.xml
@@ -1,0 +1,63 @@
+<domain type="kvm">
+  <name>vm1</name>
+  <uuid>00000000-1111-2222-3333-444444444444</uuid>
+  <memory>65536</memory>
+  <currentMemory>65536</currentMemory>
+  <vcpu>1</vcpu>
+  <os firmware="efi">
+    <type arch="x86_64" machine="pc-i440fx-6.1">hvm</type>
+    <firmware>
+      <feature enabled="yes" name="enrolled-keys"/>
+    </firmware>
+  </os>
+  <features>
+    <acpi/>
+    <apic/>
+    <vmport state="off"/>
+  </features>
+  <cpu mode="host-passthrough"/>
+  <clock offset="utc">
+    <timer name="rtc" tickpolicy="catchup"/>
+    <timer name="pit" tickpolicy="delay"/>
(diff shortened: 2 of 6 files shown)


crobinso:
I'm having second thoughts about the naming here, having `secure-boot=off` map to `enrolled-keys=off`, when there's also a `secure-boot` feature in the XML. I agree `enrolled-keys=off` is the better XML change here but the CLI naming kinda paints us into a corner that we can't get out of once it ships, without coming up with new CLI naming that diverges from libvirt + qemu naming conventions

 I'm thinking we should have both `secure-boot=on|off` and `enrolled-keys=on|off` and use man page to point people to enrolled-keys as the preferred option. If you agree I can send a patch.

phrdina:
> I'm having second thoughts about the naming here, having `secure-boot=off` map to `enrolled-keys=off`, when there's also a `secure-boot` feature in the XML. I agree `enrolled-keys=off` is the better XML change here but the CLI naming kinda paints us into a corner that we can't get out of once it ships, without coming up with new CLI naming that diverges from libvirt + qemu naming conventions
> 
> I'm thinking we should have both `secure-boot=on|off` and `enrolled-keys=on|off` and use man page to point people to enrolled-keys as the preferred option. If you agree I can send a patch.

I was thinking the same but main argument for adding this option is to make it easy for users to disable secure boot. Introducing both `secure-boot` and `enrolled-keys` would be IMO still confusing and would basically be short versions of the firmware[].features.

To me this new option is a single knob to enable/disable secure-boot in order to boot OS without proper signature. If in future some other combination of libvirt + {hypervisor} uses different XML elements we can make sure virt-install will make correct XML changes.

crobinso:
> I was thinking the same but main argument for adding this option is to make it easy for users to disable secure boot. Introducing both `secure-boot` and `enrolled-keys` would be IMO still confusing and would basically be short versions of the firmware[].features.
> 
> To me this new option is a single knob to enable/disable secure-boot in order to boot OS without proper signature. If in future some other combination of libvirt + {hypervisor} uses different XML elements we can make sure virt-install will make correct XML changes.

OK I see what you mean. I think my main fear is the conflict with the `secure-boot` name but I think we can avoid that satisfactorily enough by adding convenience `firmware.secure-boot=on|off` and `firmware.enrolled-keys=on|off` so there's a nicer CLI way to get finegrained access to the feature flags. That makes it a bit more obvious that `secure-boot` is a special option and not just a proxy for the firmware feature flag.

phrdina:
> > I was thinking the same but main argument for adding this option is to make it easy for users to disable secure boot. Introducing both `secure-boot` and `enrolled-keys` would be IMO still confusing and would basically be short versions of the firmware[].features.
> > To me this new option is a single knob to enable/disable secure-boot in order to boot OS without proper signature. If in future some other combination of libvirt + {hypervisor} uses different XML elements we can make sure virt-install will make correct XML changes.
> 
> OK I see what you mean. I think my main fear is the conflict with the `secure-boot` name but I think we can avoid that satisfactorily enough by adding convenience `firmware.secure-boot=on|off` and `firmware.enrolled-keys=on|off` so there's a nicer CLI way to get finegrained access to the feature flags. That makes it a bit more obvious that `secure-boot` is a special option and not just a proxy for the firmware feature flag.



> > I was thinking the same but main argument for adding this option is to make it easy for users to disable secure boot. Introducing both `secure-boot` and `enrolled-keys` would be IMO still confusing and would basically be short versions of the firmware[].features.
> > To me this new option is a single knob to enable/disable secure-boot in order to boot OS without proper signature. If in future some other combination of libvirt + {hypervisor} uses different XML elements we can make sure virt-install will make correct XML changes.
> 
> OK I see what you mean. I think my main fear is the conflict with the `secure-boot` name but I think we can avoid that satisfactorily enough by adding convenience `firmware.secure-boot=on|off` and `firmware.enrolled-keys=on|off` so there's a nicer CLI way to get finegrained access to the feature flags. That makes it a bit more obvious that `secure-boot` is a special option and not just a proxy for the firmware feature flag.

That sounds reasonable to introduce these new options.

I'll update the man page to drop the last sentence `This is a convenience option to control the enrolled-keys firmware feature.` to make `secure-boot=on|off` more generic.

PR #1055, opened 2026-03-01
Implement OVA (VMWare, VirtualBox) import and auto configuration.

SeraVault:
This commit adds full OVA import support to the GUI.  An additional option under "New VM" (import OVA) allows the user to import an OVA that was exported from VMWare or VirtualBox easily.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 *~
 *.pyc
+__pycache__/
 *.gmo
 
 /build
--- a/ui/createvm.ui
+++ b/ui/createvm.ui
@@ -328,6 +328,24 @@
                                         <property name="position">3</property>
                                       </packing>
                                     </child>
+                                    <child>
+                                      <object class="GtkRadioButton" id="method-ova">
+                                        <property name="label" translatable="yes">Import OV_A archive (VMware, VirtualBox)</property>
+                                        <property name="visible">True</property>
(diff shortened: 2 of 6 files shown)


phrdina:
Hi, thanks for the contribution. I did check the changes quickly but before I dive into a full review there are some things that need to be sorted.

1. This is a big change for a single commit, I think it should be possible to split it logically into several commits.
2. There is already existing tool called `virt-v2v` that already handles importing VMs between different hypervisors so I don't think we should re-invent the logic in virt-manager.

SeraVault:
I agree we shouldn't reinvent a wheel; however, these command line tools scare people.  I thought it would be nice to include this as an option in the GUI for everyone.  I agree it's a big change but it's all necessary in order to implement the functionality.  If you'd like it split could you provide some suggestions?

phrdina:
There are multiple design flaws with the code in this PR:

- The changes in `osdict.py` basically duplicates the purpose of libosinfo, the proper way would be to extend libosinfo to detect OS from the OVA image
- it hard-codes a lot of default values like memory, vcpus and so on instead of using existing virt-manager logic
- it duplicates already existing logic in virt-manager, for example creating storage volume
- as pointed out it reinvents what virt-v2v already does.

Closing this PR as IMO this doesn't belong to virt-manager.

PR #1057, opened 2026-03-03
Detect optional is_recommended attribute in the osinfo-db firmware

charlesa:
The is_recommended firmware attribute in the osinfo database allows an OS to declare efi as preferred for its security features while still supporting legacy bios. This patch checks for the attribute and defaults to the recommended firmware if specified.


charlesa:
Missed a merge with this PR. Starting over.

PR #1063, opened 2026-03-15
serialcon: add 'nmdm' console support

novel:
The 'nmdm' is a console type used by the bhyve driver. It works like a pty console, so adding support is just a matter of adding it to the 'usable_types' list.


novel:
A quick test: https://people.freebsd.org/~novel/misc/virt-manager-nmdm-console.gif

PR #1064, opened 2026-03-15
VM grouping

al4r0:
Store VM group assignments in ~/.config/virt-manager/vm-groups.json keyed by VM UUID. 
Adds group header rows, Assign Group button.


phrdina:
Hi, thanks for the contribution but as per our https://github.com/virt-manager/virt-manager/blob/main/DESIGN.md#previously-rejectedremoved-ui-elements document this was already rejected. I understand that this may be useful for some users but the design decision still stands.

PR #1073, opened 2026-03-31
[wontfix] Add --qt flag for PyQt6 frontend

kairosci:
## Summary

Add a complete Qt/KDE frontend for virt-manager accessible via `--qt` flag, while keeping the existing GTK frontend as default.

## Changes

- Added `--qt` flag to launch PyQt6 frontend instead of GTK
- Implemented complete Qt frontend in `qt/` directory with:
  - Core engine, config, signals, and connection manager
  - Bridge layer wrapping libvirt for Qt compatibility
  - Models for Connection, VM, and Storage
  - Full UI with MainWindow, dialogs, details panel, console
  - Internationalization support via gettext

## Implementation Details

- All functions that exist in GTK are implemented in Qt (no placeholders, no workarounds)
- Core `virtinst/` remains unchanged for upstream compatibility
- GTK frontend NOT replaced - both frontends coexist

## Files Changed

- 56 files, ~8900 insertions
- `qt/` - Complete Qt frontend implementation
- `virt-manager` - Entry point with --qt flag

## Testing

```bash
./virt-manager --qt  
```

## Motivation

Many users prefer Qt/KDE over GTK, and having a native Qt frontend improves integration with KDE Plasma desktop environments.


phrdina:
Hi, I guess LLM was used to write or help with writing the code and this PR, you should add that detail into commit message.

The code is definitely not production ready, for example running `./virt-manager-qt` doesn't even work, the UI looks different and doesn't seem to implement everything from GTK so that claim in this PR is not correct.

Now before I even start with proper review we need to have a discussion if we even want to support Qt in virt-manager and if yes how it should be handled.

I understand that for KDE users it may be appealing to have Qt based virt-manager GUI but it comes with a lot of extra work to make sure both GTK and Qt offer the same set of features or they would offer different set of features which to me doesn't sound like a good idea.

I'm not convinced this is something we want to do in virt-manager, at least not now.

kairosci:
Right command is `virt-manager --qt`

PR #1106, opened 2026-06-15
virtinst: fixes for iommufd - CLI parsing and CLI test update

NathanChenNVIDIA:
This PR introduces two fixes for iommufd support in virtinst:

- driver.iommufd=on emits invalid-looking iommufd="on" instead of iommufd="yes". Add is_onoff=True to cli.py to address this.

- Separate iommufd test out of many-devices test to avoid Libvirt version mismatch. Gate iommufd test behind Libvirt 12.1.0.

Testing:

- driver.iommufd=on emits iommufd="yes"

### DIFF
--- a/tests/data/cli/compare/virt-install-iommufd.xml
+++ b/tests/data/cli/compare/virt-install-iommufd.xml
@@ -1,0 +1,88 @@
+<domain type="kvm">
+  <name>fedora</name>
+  <uuid>00000000-1111-2222-3333-444444444444</uuid>
+  <metadata>
+    <libosinfo:libosinfo xmlns:libosinfo="http://libosinfo.org/xmlns/libvirt/domain/1.0">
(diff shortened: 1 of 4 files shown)


crobinso:
Not sure if it's this PR specifically, but there's test failures now running against libvirt.git, I reproduced locally as well: https://github.com/virt-manager/virt-manager/actions/runs/33834224604/job/100903300529

There's a typo in test_cli.py: s/driver.ssidSize/driver.ssidsize/ , but maybe there's more to it

phrdina:
Thanks for the report, I missed that part, libvirt changed the XML attribute to `ssidsize` before it was merged but this test used the old `ssidSize` so that needs to be fixed and also boot order in the output XML for iommufd test was incorrect.

PR fixing both issues: https://github.com/virt-manager/virt-manager/pull/1140

PR #1108, opened 2026-06-16
Add bhyve as a default hypervisor

jainnimish:
This patch allows virt-manager to detect bhyve as the default hypervisor on FreeBSD hosts.

### DIFF
--- a/virtManager/createconn.py
+++ b/virtManager/createconn.py
@@ -37,6 +37,9 @@ def _default_uri():  # pragma: no cover
         "/usr/lib64/libvirt/libvirt_lxc"
     ):
         return "lxc:///"
+
+    if os.path.exists("/usr/sbin/bhyve"):
+        return "bhyve:///system"
     return None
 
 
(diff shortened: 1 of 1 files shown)


ptoscano:
You probably want to tweak also the `set_default_hypervisor()` function in that file.

jainnimish:
I have done so.

virtManager/createconn.py
```diff
@@ -37,6 +37,9 @@ def _default_uri():  # pragma: no cover
         "/usr/lib64/libvirt/libvirt_lxc"
     ):
         return "lxc:///"
+
+    if os.path.exists("/usr/sbin/bhyve"):
+        return "bhyve:///system"
```
novel, line 42: At first I was going to suggest to place that before the qemu checks so we match bhyve first even if qemu is installed on the system. But then realized that we will actually never match qemu installed from ports/packages because it's installed in the `/usr/local/` prefix. We should probably fix it at some point, but that looks like a very minor issue.

novel:
Tested the change, works fine for me.
Code looks good, but I think it needs to be squashed into a single commit.

virtManager/createconn.py
jainnimish: Hmm, yes.

jainnimish:
> Tested the change, works fine for me. Code looks good, but I think it needs to be squashed into a single commit.

Squashed.

novel:
ping?

Issue #142, opened 2020-08-23
RFE: more snapshot sorting options: creation time, snapshot hierarchy


bergentroll:
It is continue of https://bugzilla.redhat.com/show_bug.cgi?id=1859015. I am sorry for duplication, I have troubles with getting access to Bugzilla.

Briefly: snapshots on details window sorted now by name. Creation time seems more reasonable target to sort.

It seems it needs very small changes to do so: the `sortname` should include `creationTime` string instead of `name`. It takes beginning here: https://github.com/virt-manager/virt-manager/blob/a121c943ef29b30c2c9cf49a79cb359c9c3cb68f/virtManager/details/snapshots.py#L455

Also concatenation of digit that is describing type of item with name is really laconic, but seems a bit fragile.

crobinso:
Thanks for the suggestion. I'm folding some other ideas into this as well, like an option to show the snapshot hierarchy as well, since the work will be in the same area. This was previously discussed at this older bug: https://bugzilla.redhat.com/show_bug.cgi?id=1065077

FWIW I doubt I will change anything here due to time constraints, until external snapshots are at a point that virt-manager can support them, at which point it will require a larger rethink of the UI which can include these improvements

jimkring:
I assumed, wrongly, that snapshots were sorted chronologically. I spent a while trying to figure out if there was a way to change this (e.g. in the `View>` menu). It feels a little bit like great is the enemy of good, with respect to having a tree view be required for sorting by date. Maybe there could be a simple toggle between sort by name vs by date?

crobinso:
external snapshot is here these days, so it's worth investigating what tweaking the snapshot list will look like

PR #1115, opened 2026-06-29
Display snapshots as a tree sorted by creation date

cfconrad:
As a daily `virt-manager` user who has recently been heavily relying on snapshots, I realized it is quite hard to see which snapshots are parents or children of others. Thus I changed the snapshot list to a tree view ordered by creation date, which has helped me a lot when managing (deleting, reverting, branching) my snapshots.

### The commit include:
- Switch `GtkListStore` to `GtkTreeStore` to render parent/child snapshot relationships
- Sort snapshots by creationTime for chronological sibling ordering
- Label both internal and external snapshots with their type
- Auto-size the paned divider to the tree's natural width on first populate
- Drop the internal/external separator row, now redundant with the tree structure
- Preserve tree expansion state across refresh, while always expanding the currently in-use snapshot

### Example: 
Screenshot_2026-06-29_17:34:29

### Related issues:
 * https://github.com/virt-manager/virt-manager/issues/142
 * https://github.com/virt-manager/virt-manager/issues/847

### DIFF
--- a/ui/snapshots.ui
+++ b/ui/snapshots.ui
@@ -32,6 +32,7 @@
                     <property name="visible">True</property>
                     <property name="can-focus">True</property>
                     <property name="headers-visible">False</property>
+                    <property name="enable-tree-lines">True</property>
                     <signal name="button-press-event" handler="on_snapshot_list_button_press_event" swapped="no"/>
                     <signal name="row-activated" handler="on_snapshot_list_row_activated" swapped="no"/>
                     <child internal-child="selection">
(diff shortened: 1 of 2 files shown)


cfconrad:
Just realized that there is already one open with similar approach: https://github.com/virt-manager/virt-manager/pull/994

I keep mine open as well, because it has some more features (like expand handling and with resize). Let me know if we do not want to keep this duplicate, thanks!

cfconrad:
Interested `openSUSE` users can use builds from my home https://build.opensuse.org/package/show/home:cfconrad:branches:Virtualization/virt-manager

PR #1122, opened 2026-07-16
xmletree: fix serialization of default (unprefixed) XML namespaces

ryan-ronnander:
When working with [virter](https://github.com/LINBIT/virter) that stores its own metadata in libvirt XML entries, I found my own system (openSUSE Tumbleweed, no `python3-libxml2` package available) failing on both  `virt-xml --edit` and `virt-install --reinstall`.

To my surprise, the root cause turned out to be here in `xmletree.py`, not in virter, just a couple lines of code to patch. The test suite passes with the new fix.

This only affects the `xmletree.py` backend, used as a fallback when the libxml2 Python bindings aren't available. Verified the `xmllibxml2.py` backend doesn't have this bug, it delegates straight to libxml2, which handles it correctly.

### DIFF
--- a/virtinst/xmletree.py
+++ b/virtinst/xmletree.py
@@ -97,7 +97,8 @@ def _convert_qname(tag, namespaces):
         uri, tag = tag[1:].rsplit("}", 1)
         for key, val in namespaces.items():
             if uri == val:
-                tag = key + ":" + tag
+                if key:
+                    tag = key + ":" + tag
                 break
     return tag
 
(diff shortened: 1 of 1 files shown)


ryan-ronnander:
Circling back to this one. I just opened [bsc#1282618](https://bugzilla.opensuse.org/show_bug.cgi?id=1282618) in the openSUSE Bugzilla, since it affects Tumbleweed. The fix still applies cleanly against current main.

PR #1131, opened 2026-08-14
devices: disk: replace str.removeprefix with startswith for Python < 3.9 compat

lienze:
str.removeprefix() was introduced in Python 3.9, but the project documents python >= 3.4 as the minimum requirement in INSTALL.md.  When virt-manager is run on Python 3.8 or older, setting the install media path via a file:// URI (E.g., from a file manager) fails with:

  'str' object has no attribute 'removeprefix'

Replace it with startswith() + slicing, which is equivalent and compatible with all supported Python versions.

### DIFF
--- a/virtinst/devices/disk.py
+++ b/virtinst/devices/disk.py
@@ -694,8 +694,8 @@ class DeviceDisk(Device):
     def set_source_path(self, newpath):
         # Some file managers use 'file://' when passing files to
         # virt-manager, we need to strip it from the newpath.
-        if newpath is not None:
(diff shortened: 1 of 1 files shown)


phrdina:
Hi, thanks for the contribution but the documented supported python version is outdated and should be changed to 3.9.

lienze:
@phrdina Thanks for the review!  Agreed -- I'll submit a new PR that updates the documented minimum Python version to 3.9 (INSTALL.md and the buildconfig runtime check) instead. :)

PR #1136, opened 2026-08-18
manager: Add groups to the VM list

Ramil66:
Allow virtual machines to be organized into groups in the manager VM list.

VMs can be moved into new or existing groups, removed from groups, and groups can be renamed or ungrouped. Groups also provide actions to start or shut down their member VMs.

Group membership is stored per VM. Ungrouped VMs continue to appear directly under their connection as before.


mikhailnov:
Looks useful

